PR #4, opened 2017-10-09
wrapper to load a structure file without instantiating a reader

danijoo:
This adds a method that allows to read any supported file format without explicitly instantiating a reader class.

```python
> from biopython.structure.io import read_structure_from_file
> read_structure_from_file('1bl8.pdb', format='pdb') # format is optional
'AtomArray'
```

### DIFF
--- a/src/biopython/structure/io/__init__.py
+++ b/src/biopython/structure/io/__init__.py
@@ -8,3 +8,4 @@ A subpackage for reading and writing structure related data.
 """
 
 from .trajfile import *
+from .structfile import *
--- a/src/biopython/structure/io/structfile.py
+++ b/src/biopython/structure/io/structfile.py
@@ -26,7 +26,7 @@ def read_structure_from_file(file_name, format=None):
     Examples
     --------
     Load a `\*.pdb` file:
-    
+        >>> from biopython.structure.io import read_structure_from_file
         >>> struct = read_structure_from_file('1l2y.pdb', format='pdb')
     """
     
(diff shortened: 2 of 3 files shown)


padix-key:
Hi,
I think the automatic reader class determination as done in

    class_path = "biopython.structure.io.{format}.{prefix}File".format(format=format.lower(), prefix=format.upper())

is too unflexible to cover all cases. For example the common file suffix for PDBx/mmCIF files is '.cif', while the reader class is called `PDBxFile`. The same is true for '.npz' and `NpzFile`. Furthermore the
`PDBxFile` class only provides dictionary like access, the corresponding `get_structure()` function is
in the same subpackage but not part of the class. Therefore I think it would be better to handle each suffix as separate case.

Furthermore the documentation for the return value is not updated yet.

danijoo:
you are right. Im not sure how to easily solve this though. The only other possibility I can think of is doing a huge if/else stack (that has to be extended with every new format). Or one could make separate wrapper classes for every fileformat. I.e:

```python
from biopython.structure.io.pdb import read_pdb
struct = read_pdb("my_pdb.pdb")

from biopython.structure.io.tng import read_tng
struct = read_pdb_to_structure("my_pdb.pdb") 

# and so on

```

padix-key:
I slightly prefer the if/else stack, since you have a central loading function then. The `read_structure_from_file` needs the optional parameter `template`in this case, in order to support trajectory files

PR #38, opened 2018-07-19
resolves #37 by guessing element names

### LINKED ISSUE #37: PDB element symbol should be guessed from atom name if not present
Strictly speaking, the elemet in a PDB structure (column 77-78) is part of the format specification and should always be set. Nevertheless, a lot of programs ignore this column and omit it in the output or only write it for backbone atoms leading to this column be only partially filled (happened to us). 

Biotite on the other hand, requires this column for various functions including selection of atoms by their types which leads to an Error when these functions are called. The circumstance that this column is often not set can be worked arround by trying to guess the element from the the atom column (line 13-16), possibly in combination with a warning to the user. 
If you are interested in a solution like that, I can make a pull request for this.

danijoo:
This should resolve issue #37. Elements are guessed from atom names by using the first character as element name if element column is empty. This should work for most protein and RNA simulation stuff but might break for exotic hetero atoms. Therefore, a warning is issued to let users know that some element names might be off.


padix-key:
It seems like I accidentally pushed some other recent commits of mine to your fork. Since these commits are already merged into biotite-dev/biotite, this shouldn't be a problem.

padix-key:
If you have no objection to my changes, I would like to merge the PR

danijoo:
looks good!

padix-key:
PR is merged. Thank you for your contribution!

PR #40, opened 2018-08-03
add support for the .gro file format

danijoo:
This pull requests adds support for the gromacs .gro file format. reading and writing structures works and are unit tested. One issue that might need to be addressed in the future is that .gro has no column for element type. Therefore, these are guessed based on the atom_name (first char of the string).


padix-key:
The failed test is not related to your PR, as it occured in `biotite.database.entrez.fetch()`.
This is maybe caused by a transient error in the NCBI Entrez server.

padix-key:
Looks good. How do .gro files handle insertion codes and altlocs?

padix-key:
I merged the PR. Thank you for the code!

PR #43, opened 2018-08-08
Detection of hydrogen bonds

danijoo:


padix-key:
I changed one function name from `get_hbond_frequency()` to `hbond_frequency()` for more consistency with `hbond()`.

padix-key:
I changed the return value of `hbond()`, so that it only returns the mask when multiple models are given, as for a single model the mask always contains only true values.

padix-key:
I will incorporate the use of `BondList` objects for donor-hydrogen detection, when a way for automatic bond detection is developed. Hence, I am going to merge this PR now. Thank you for your effort.

PR #52, opened 2018-08-22
allow string/file-like as template for load_structure

danijoo:
Right now, loading a trajectory is a two-step because the template has to be loaded in memory before calling ```load_structure```. This can be reduced to a oneliner by allowing to submit strings as templates:

```python
from biotite.structure.io import load_structure

# "abc.pdb" is automatically loaded as template 
stack = load_structure("abc.xtc", template="abc.pdb")
```

### DIFF
--- a/src/biotite/structure/io/general.py
+++ b/src/biotite/structure/io/general.py
@@ -27,7 +27,7 @@ def load_structure(file_path, template=None):
     ----------
     file_path : str
         The path to structure file.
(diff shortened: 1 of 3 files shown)


padix-key:
Good idea. I modified the type check for the `template` parameter, so that it directly checks for a string or file-like object. Furthermore, I added some type annotations.

padix-key:

PR #78, opened 2019-01-28
Fix geometry.distance documentation

danijoo:
The doc for the "box" parameter in geometry.distance happend to end in the Returns section instead of Parameters leading to wrong autocomplete suggestions.


padix-key:
Thank you for the fix

PR #99, opened 2019-04-10
handle dummy boxes in .gro correctly

danijoo:
There is a small issue with the .gro input reader resulting in an exception if an empty box (x=y=z=0.000) is read in. This fixes this problem and also implements an appropriate unit test.


padix-key:

Issue #101, opened 2019-04-17
backbone dihedral calculation without chain ids


danijoo:
The current function parameters for ```biotite.structure.dihedral_backbone``` require to supply a chain id to calculate backbone torsion angles. Because some file formats (.gro) do not support chain ids, this can make it a bit cumberstone to use this function. What is your opinion on making ```chain_id``` an optional parameter and calculate dihedrals on the whole protein if ```chain_id``` is ```None```? This would also make it a lot easier to calculate dihedrals for multimers because we can call the function once to get dihedrals for all chains instead of calling it multiple times. 

```python
def dihedral_backbone(atom_array, chain_id=None):
    do stuff
```

danijoo:
It should be a simple change to make and I can do a PR if you are interested. Just wanted to bring this to discussion first :)

padix-key:
Sounds like a good idea. Since I wanted to make some tweaks to this function anyway, I will incorporate the optional parameters in this process.

padix-key:
The recent PR solved this problem by completely removing this parameter.

PR #102, opened 2019-04-17
Revamped 'dihedral_bockbone()' function

padix-key:
This PR changes the `dihedral_backbone()` function:

- Added chain-level functionality, similar to the existing residue-level functionality
- The `chain_id`, parameter is removed, the function iterates over all chains
- Improved validity check for input atom arrays


padix-key:
Fixes #101

PR #109, opened 2019-05-24
dihedrals: correct output shape for multiple chains

danijoo:
The result for multiple chains is concatenated along the wrong axis. For X frames, Y residues and Z chains, the result should be of the shape (X, Y*Z), but the current outbut is (X*Z, Y). This resolves this issue.


padix-key:
Your approach works for atom array stacks objects but fails for single atom arrays, since their dihedral angles only have one axis. I changed it, so the joining is always performed over the last axis.

Issue #112, opened 2019-06-23
Add support for hybrid serial numbers in PDB format?


danijoo:
The current limitation of the pdb format is that it only supports serial numbers up to 99999 atoms and 9999 residues. An enhanced version of the pdb format called "hybrid-36 pdb" overcomes this by switching to a hex representation for those ids. 

Example
base pdb format
```
ATOM      1  O           1       0.251  -0.360  -0.046  1.00  0.00
ATOM      2  H           1       0.249   0.684   0.231  1.00  0.00
ATOM      3  H           1       0.586  -0.954   0.791  1.00  0.00
...
ATOM  99997  O           1       0.251  -0.360  -0.046  1.00  0.00
ATOM  99998  H           1       0.249   0.684   0.231  1.00  0.00
ATOM  99999  H           1       0.586  -0.954   0.791  1.00  0.00
ATOM      1  O           1       0.251  -0.360  -0.046  1.00  0.00
ATOM      2  H           1       0.249   0.684   0.231  1.00  0.00
ATOM      3  H           1       0.586  -0.954   0.791  1.00  0.00
```

hybrid-36 pdb
```
ATOM      1  O           1       0.251  -0.360  -0.046  1.00  0.00
ATOM      2  H           1       0.249   0.684   0.231  1.00  0.00
ATOM      3  H           1       0.586  -0.954   0.791  1.00  0.00
...
ATOM  99997  O           1       0.251  -0.360  -0.046  1.00  0.00
ATOM  99998  H           1       0.249   0.684   0.231  1.00  0.00
ATOM  99999  H           1       0.586  -0.954   0.791  1.00  0.00
ATOM  A0000  O           1       0.251  -0.360  -0.046  1.00  0.00
ATOM  A0001  H           1       0.249   0.684   0.231  1.00  0.00
ATOM  A0002  H           1       0.586  -0.954   0.791  1.00  0.00
```

As you can see, the big advantage of this format is that it keeps backward compatibility if the residue and serial id are below the threshold for hex numbers - which is a clear advantage over other efforts to overcome the limitations of pdb.

This format was recently implemented in Modeller 9.22 and is also available an various other software packages like MDAnalysis. I think an implementation of this for biotite would not require a lot of code changes: The pdb reader can do this internally when parsing the serial and residue field. And for output, it could be either a modified version of PDBFile or a simple flag to alter the output numbering.

I'd like to put this up for diskussion here before making the effort to implement it. What's your opinion? Yay or nay ;)

[...]

PR #126, opened 2019-08-25
Hybrid-36 support for PDB

danijoo:
Here is an initial implementation that extends the pdb format with hybrid-36 (see #112). It adds a flag to `set_structure()` to write as hybrid-36 and automatically detects/parses hybrid-36 in `get_structure()`. It uses the suggested reference implementation from [cctbx](https://github.com/cctbx/cctbx_project/blob/master/iotbx/pdb/hybrid_36.py).

```python
length = 100005
a = AtomArray(length)
a.coord = np.zeros(a.coord.shape)
a.chain_id = np.full(length, "A")
a.res_id = np.arange(1, length+1)  # Create residue IDs over 10000
a.res_name = np.full(length, "GLY")
a.hetero = np.full(length, False)
a.atom_name = np.full(length, "CA")
a.element = np.full(length, "C")

f = pdb.PDBFile()
f.set_structure(a, hybrid36=True) 
f.write("/tmp/hybrid.pdb")
```

```bash
> tail /tmp/hybrid.pdb
ATOM  99996 CA   GLY ABXFW       0.000   0.000   0.000  1.00 0.000          C   
ATOM  99997 CA   GLY ABXFX       0.000   0.000   0.000  1.00 0.000          C   
ATOM  99998 CA   GLY ABXFY       0.000   0.000   0.000  1.00 0.000          C   
ATOM  99999 CA   GLY ABXFZ       0.000   0.000   0.000  1.00 0.000          C   
ATOM  A0000 CA   GLY ABXG0       0.000   0.000   0.000  1.00 0.000          C   
ATOM  A0001 CA   GLY ABXG1       0.000   0.000   0.000  1.00 0.000          C   
ATOM  A0002 CA   GLY ABXG2       0.000   0.000   0.000  1.00 0.000          C   
ATOM  A0003 CA   GLY ABXG3       0.000   0.000   0.000  1.00 0.000          C   
ATOM  A0004 CA   GLY ABXG4       0.000   0.000   0.000  1.00 0.000          C   
ATOM  A0005 CA   GLY ABXG5       0.000   0.000   0.000  1.00 0.000          C 
```

What needs to be addressed:
- [ ] Benchmarking: If this makes the pdb reader significantly slower, we need to see if there is a better way to handle the hy36 detection. Right now, everything is thrown into the hy36decode function which works for hy36 as well as base-10. Thats probably not very efficient and we should see address this. Probably with cython.
- [ ] We should consider reimplementing [hybrid_36.py](https://github.com/danijoo/biotite/blob/hybrid36/src/biotite/structure/io/pdb/hybrid_36.py) for better performance and to match biotite coding conventions.

Opinions?


danijoo:
failing tests are doctest - not part of the PR

padix-key:
The problem is that the `__all__` attribute is missing from `hybrid_36.py`. Could you add it?

padix-key:
Does the source code for the hybid36 encoding require a license?

danijoo:
The hybrid 36 source code contains ```This file is unrestricted Open Source (cctbx.sf.net).``` However, the origin repo has this [license](https://github.com/cctbx/cctbx_project/blob/master/LICENSE.txt), which mentions that ```(1) Redistributions of source code must retain the above copyright
notice, this list of conditions and the following disclaimer.```.

I'm not sure what this means for us. Unfortunately, cctbx us not on pypy/conda, so we cannot use it as a dependency. We can either somehow incorporate their license agreement in the hybrid_36.py and link to it, or write our own version.

padix-key:
I added a C-accelerated version of the hybrid-36 codec, resolving the license issue. While the encoding is ~6 times faster than in the reference implementation, the decoding is surprisingly only ~ 50% faster.

padix-key:
I think all issues should be resolved now. If you agree, I would like to merge the PR.

danijoo:
Lets merge it!

PR #138, opened 2019-09-15
Allow passing additional keyword arguments to io.load_structure method

danijoo:
Keyword arguments submitted as **kwargs will be passed to the files `get_structure` method for non-trajectories (pdb, gro..) and to `read` for trajectories (xtc, tng, ....). It seems like no file format has additional arguments for both so far.

The only weak spot is template loading from a string, which won't get additional arguments. This would require creating (and maintaining) a list of allowed arguments for `read` and `get_structure` for every possible trajectory/template combination. Users can pass AtomArrays instead if it is required to pass additional arguments to a template.

### DIFF
--- a/src/biotite/structure/io/general.py
+++ b/src/biotite/structure/io/general.py
@@ -15,7 +15,7 @@ import io
 from ..atoms import AtomArray, AtomArrayStack
 
 
-def load_structure(file_path, template=None):
+def load_structure(file_path, template=None, **kwargs):
     """
     Load an atom array or stack from a structure file without the need
     to manually instantiate a :class:`File` object.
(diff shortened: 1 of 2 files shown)


padix-key:
Thank you, this looks like a sensible enhancement of the function. I would like to ask for two changes/additions:

- Could you also give the **kwargs to `set_structure()` of the `NpzFile`? Although NpzFile does not accept kwargs it would ensure an exception is raised if a user tries regardlessly.
- Could you implement the same behavior for `save_structure()`? In the moment this would make only sense for PDBx files, but this may change in the future.

danijoo:
there you go!

PR #151, opened 2019-11-06
add shape paramenter

danijoo:
Adds an additional property to AtomArrays and Stacks similar to np.shape.


padix-key:
Thank you. I added `shape` to the object attributes in the class docstrings.

PR #152, opened 2019-11-06
function to calculate atomistic densities

danijoo:
This PR adds a function to perform partial density calculation, i.e. to calculate the probability distribution of a molecule in a simulation box.

In the current state, this PR should be considered as a "draft". There are some things we need to consider:

- [x] unit tests
- [x] Weights: we can simply forward a weight parameter to `np.histogramdd` or do something more sophisticated. This is required to weight the density based on atomic parameters like the mass or electron density.
- [ ] Can we easily export densities to disk, i.e. in .eds file format? I think this is widely used for electron density maps from cryo EM. It would be nice to be able to compare calculated densities with experimental data.

### DIFF
--- a/src/biotite/structure/__init__.py
+++ b/src/biotite/structure/__init__.py
@@ -101,6 +101,7 @@ from .bonds import *
 from .box import *
 from .celllist import *
 from .compare import *
+from .density import *
 from .error import *
 from .filter import *
 from .geometry import *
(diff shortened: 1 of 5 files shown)


danijoo:
weights and unit tests are done. I think point 3 in the above list is very optional and we can address this later if we want.

padix-key:
I agree. I think direct support for binary `.eds` files would be too special for *Biotite's* scope.

padix-key:

PR #164, opened 2019-11-26
pdb raise ValueError for unknown extra fields. fixes #162

### LINKED ISSUE #162: pdb.get_structure should throw an error for unknown extra fields
Hi,

```Python
import biotite.structure.io as io
s = io.load_structure("1l2y.pdb", extra_fields="asdasd")
s.asdasd
> AttributeError: 'AtomArray' object has no attribute 'asdasd'
```

In the above example, `load_structure` does not throw an error for the passed extra_field even though its not a valid one. I think an error thrown during load_structure would be a nicer way to handle this because the caller immediately knows something went wrong.\

danijoo:


padix-key:
Thank you for the fix!

PR #199, opened 2020-05-06
Better element guesses (closes #188)

### LINKED ISSUE #188: Structure submodule: Replace errorprone element detection
A lot of programs do not write element names in pdb files. Gro files miss that column entirely. Right now, biotite tries to "guess" the element for these rows based on the atom name. However, due to the large varity of force fields and programs out there, this is very limited and fails often, i.e when:

- atom name does not match the element name (1H, 2H)
- Correct element abbreviation has more than one letter
- For virtual sites in general (gromacs)

I suggest to limit the element detection to cases where the element can be predicted accurately (i.e "CA" almost always matches to element="C", "H*" and "[0-3]H" being "H" and  "O*" is "O" is shared among almost every biological software). In less obvious cases, I suggest to place an empty string instead.

danijoo:


padix-key:
I added some more element symbols to the test cases.

padix-key:

PR #210, opened 2020-06-03
Id check

danijoo:
Here are some convenient functions I tend to use quite often on broken pdb files.

1) I noticed there is a function to check for gaps in residue ids, but not for atom ids. I added a method for this and renamed the other one so they are not ambigeous: `check_atom_id_continuity` and `check_res_id_continuity`

2) After a failed integrity check, you usually want to fix the errors in the array. I added two methods that allow to renumber the residues and atom ids which removes all gaps, but keep similar resids: `renumber_atom_ids` and `renumber_res_ids`:

i.e:
atom ids `[1,2,3,4,6,7]` would become `[1,2,3,4,5,6]` (yes, this is trivial)
residue ids `[1,1,2,2,4,4,5,4]` would become `[1,1,2,2,3,3,4,5]`

### DIFF
--- a/doc/apidoc.json
+++ b/doc/apidoc.json
@@ -228,6 +228,8 @@
         "Checks" : [
             "check_bond_continuity",
             "check_id_continuity",
+            "check_atom_id_continuity",
+            "check_res_id_continuity",
             "check_duplicate_atoms"
         ],
         "Residue level utility" : [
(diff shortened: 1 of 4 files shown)


padix-key:
Is there a reason for using `np.hstack()` instead of `np.concatenate()` in `integrity.py`?

danijoo:
No. they are more or less equivalent in this case, since hstack uses concatenate under the hood anyways. The solution is pretty much copy+pasta from something else, which is the reason for it being hstack in the first place :)

padix-key:
OK, then I will change this to `concatenate()`.

padix-key:
> copy+pasta

🍝

padix-key:

PR #225, opened 2020-08-06
Add 'get_molecular_weight()' function to ProteinSequence class

thomasnevolianis:
- Implementation of `get_molecular_weight` function to the `ProteinSequence` class.

- Add of` test_get_molecular_weight` to the` test_seqtypes.py`.


padix-key:
Thank you for your contribution. When I type the test sequence `ACDEFGHIKLMNPQRSTVWY`into https://web.expasy.org/compute_pi/ I get a molecular weight of 2395.74 (average) and 2394.12 (monoisotopic) instead of the 2378.11 in the test function. Maybe there are wrong values in `_mol_weight`?

padix-key:
Continued in #228.

PR #228, opened 2020-08-07
Get molecular weight average monoisotopic

thomasnevolianis:
- Add **monoisotopic** and **average** options of calculating the molecular weight of a protein to `get_molecular_weight` in `seqtypes.py`

- Add parametrization tests for the **monoisotopic** and **average** to `test_get_molecular_weight` in `test_seqtypes.py`.


padix-key:

PR #239, opened 2020-10-26
Add 'base_stacking()'

tomtomhdx:
Changed the criterion for the `_check_base_stacking` function such that the angle between one normalized distance vector and **each** of the bases' normal vectors or supplement must be <= 40° as **opposed to just one** normal vector. Intuitively this ensures that a stacking interaction is only recognized if the base planes are *parallel*.
Furthermore, the function `base_stacking` was added to check any polynucleotide for base stacking interactions using the criteria from `_check_base_stacking`, which is also utilized by `base_pairs`.


padix-key:
Thank you! There are some minor issues I would like to ask you to solve:

- There are still some tests failing, because the `.cif` and `.mmtf` formats for `1BNA` are missing. Could you please add them too? 
- Could you use the MMTF format in `test_base_stacking()`? This souls decrease the execution time of the test.
- Could you explain why you expect the stacking interactions at the given positions
- There is a typo in the variable `exspected_stackings`.

padix-key:
I made some adjustments to the docstring of the function.

padix-key:

PR #247, opened 2020-11-10
Improve 'base_stacking()'

tomtomhdx:
This PR improves the pre-selection of potential stacking partners in  `base_stacking()`. 

`_get_proximate_basepair_candidates()` was renamed to `_get_proximate_residues()`. Also the function is now more generalized:
- The selection of atoms whose distances should be compared is now specified by a boolean mask
- The default cutoff value was removed

Previously `base_stacking()` pre-selected potential stacking candidates using `_get_proximate_basepair_candidates()`, with a cutoff distance between N/O atoms pairs of 15Å. The function was originally implemented for pre-selecting potential basepair candidates for `base_pairs()` based on the plausibility of hydrogen bonds. In this case the function was used to approximate the distance between the bases. 

However, in order to approximate the bases distance one atom per base can be sufficient. In this PR the preselection of potential stacking candidates is done using only the C1'-Atom of the bases glycosidic bonds with the same cutoff of 15Å.

### DIFF
--- a/doc/apidoc.json
+++ b/doc/apidoc.json
@@ -1,4 +1,4 @@
-{   
+{
     "biotite" : {
         "File classes" : [
             "File",
@@ -13,7 +13,7 @@
             "AdaptiveFancyArrow"
         ]
     },
-    
+
     "biotite.application" : {
         "Application classes" : [
             "Application",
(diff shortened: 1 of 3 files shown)


padix-key:

PR #267, opened 2020-12-15
Fix del_annotation.py

ebetica:
Currently, it only tries to delete non-existent annotations.


padix-key:
You are totally right. Thank you very much.

PR #275, opened 2021-01-05
Improve performance of pseudoknot detection

tomtomhdx:
This PR improves the performance of `pseudoknots()`.

The helper function `_conflict_cliques()`  was adjusted to allow the separation and thus independent computation of nested cliques of conflicting regions. Previously, the function was only able to separate neighboring cliques of conflicting regions.


padix-key:
Thanks for the contribution!

PR #276, opened 2021-01-05
Fix bugs in `base_pairs()`

tomtomhdx:
This PR fixes the following bugs and adds a test function checking their presence:

- The `min_atoms_per_base` argument was only considered for non-canonical bases. This led to an exception being thrown if no atoms in a base with a canonical residue name matched the corresponding reference base.
- If no base pairs were found, an exception was thrown.


padix-key:
Could you please check for this warning also in the new test function?

tomtomhdx:
> Could you please check for this warning also in the new test function?

Done

padix-key:
Looks good to me. Thank you for your contribution!

PR #299, opened 2021-03-20
Part of issue 290

### LINKED ISSUE #290: String representation of objects
It would be nice if we could add the `__repr__` method to biotite objects such as sequences, etc. Representations are especially useful when working in Jupiter notebook.

MaxGreil:
Partly fixes #290 

- NucleotideSequence
- ProteinSequence
- GeneralSequence
- Alphabet
- AlphabetLetter

### Additional information

@padix-key I'm not quite sure how to implement method `__repr__` for abstract base class `Sequence`

### DIFF
--- a/CONTRIB.rst
+++ b/CONTRIB.rst
@@ -6,3 +6,4 @@ CONTRIBUTORS
 - Patrick Kunzmann <https://github.com/padix-key>
 - Tom David Müller <https://github.com/tomtomhdx>
 - Thomas Nevolianis <https://github.com/thomasnevolianis>
+- Maximilian Greil <https://github.com/MaxGreil>
--- a/src/biotite/sequence/alphabet.py
+++ b/src/biotite/sequence/alphabet.py
@@ -106,7 +106,7 @@ class Alphabet(object):
 
     def __repr__(self):
         """Represent Alphabet as a string for debugging."""
-        return 'Alphabet(%s)' % self._symbols
+        return f'Alphabet({self._symbols})'
 
     def get_symbols(self):
         """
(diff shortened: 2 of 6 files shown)


padix-key:

> I'm not quite sure how to implement method `__repr__` for abstract base class `Sequence`

Yes, that's true. I think I would simply omit the method for the base class then.

I would like to request a few further changes:

1. Could you use f-strings for formatting? For example, in my opinion  `f"NucleotideSequence({''.join(self.symbols)}, ambiguous={ambiguous})"` looks clearer than `'NucleotideSequence("%s", ambiguous=%s)' % (''.join(self.symbols), ambiguous)'`.
2. Could you add a test module in the `tests/`directory containing a function that tests for a selection of objects, whether `eval(repr(object)) == object`?

MaxGreil:
@padix-key I applied the changes.

Are the new tests in `tests/sequence/test_seqtypes.py ` like you had in mind?

padix-key:
I would put the test into `tests/test_repr.py` with the function name `test_repr()`, since it also can be used for classes in other subpackages of Biotite. Testing mutiple objects the same way is cleaner with `pytest.mark.parametrize` (https://docs.pytest.org/en/stable/parametrize.html): You can put the different objects into the parameterization and the function itself only executes `assert eval(repr(object)) == object`. This way pytest indicates directly which object failed, in case the `repr()` for one object fails.
This approach also requires to import the classes directly (e.g. `from biotite.sequence import NucleotideSequence` instead of import biotite.sequence as seq), to get rid of the `..."seq." + ...` part of the test function.

For the sake of completeness, could you also add a test case for an ambiguous `NucleotideSequence`?

Don't mind the curently failing test, I think this is a temporary problem with the PDB.

MaxGreil:
@padix-key I applied the latest changes

padix-key:
This looks good to me now, thank you. If you like, you may also add yourself to the `CONTRIB.rst`.

MaxGreil:
I added myself to `CONTRIB.rst`.

PR #301, opened 2021-03-28
Issue 290 Part 2

### LINKED ISSUE #290: String representation of objects

MaxGreil:
Fixes #290

New __repr__ method for classes:

- CodonTable
- Alignment
- SubstitutionMatrix
- Location
- Feature
- Annotation
- AnnotatedSequence
- Atom

Classes AtomArray and AtomArrayStack can use __repr__ method from respective superclass

### Questions

@padix-key I think for classes CodonTable and SubstitutionMatrix are new __eq__ and __ne__ methods necessary? What do you think?

### DIFF
--- a/src/biotite/structure/atoms.py
+++ b/src/biotite/structure/atoms.py
@@ -494,15 +494,13 @@ class Atom(Copyable):
 
     def __repr__(self):
         """Represent Atom as a string for debugging."""
-        annot = ''
-        for name, annotation in self._annot.items():
-            if len(annot) == 0:
-                annot = name + '="' + annotation + '"'
-            else:
-                if type(annotation).__name__ == 'str':
-                    annot = annot + ', ' + name + '="' + annotation + '"'
-                else:
-                    annot = annot + ', ' + name + '=' + str(annotation)
+        annot = 'chain_id="' + self._annot["chain_id"] + '"'
+        annot = annot + ', res_id=' + str(self._annot["res_id"])
+        annot = annot + ', ins_code="' + self._annot["ins_code"] + '"'
+        annot = annot + ', res_name="' + self._annot["res_name"] + '"'
+        annot = annot + ', hetero=' + str(self._annot["hetero"])
+        annot = annot + ', atom_name="' + self._annot["atom_name"] + '"'
+        annot = annot + ', element="' + self._annot["element"] + '"'
         return f'Atom(np.{np.array_repr(self.coord)}, {annot})'
 
     @property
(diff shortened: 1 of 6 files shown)


padix-key:
Thank you for you effort! :+1:

>  I think for classes CodonTable and SubstitutionMatrix are new eq and ne methods necessary? What do you think?

I agree. Could you add them in this PR?

> Classes AtomArray and AtomArrayStack can use repr method from respective superclass

Could you explain? The class `_AtomArrayBase` has no implemented `__repr__()` method.

For the `CodonTable` class I would like to request to put the `starts` parameter name into the repr-string. I think this makes it clearer for the user what the second parameter means.

MaxGreil:
> Could you explain? The class _AtomArrayBase has no implemented __repr__() method.

You are right @padix-key , I will also implement the `__repr__` method for classes `AtomArray` and `AtomArrayStack`. Regarding these two classes, should the __repr__` method look like in the discussion for issue #290

`array([
    Atom(array([1.0, 2.0, 3.0]), chain_id="A", res_id=1, ...),
    ...
])
`
?

padix-key:
I would say so. However, there is a name conflict, as the outer `array()` refers to `structure.array()` and the inner `array()` refers to `numpy.array()`. Maybe we change the inner function to `'np.array([...])'`? For `AtomArrayStack` I would propose something like
```
stack([
    array(...),
    array(...)
])
```

MaxGreil:
@padix-key I applied the changes

src/biotite/structure/atoms.py
```diff
@@ -660,7 +673,17 @@ def __init__(self, length):
...
+
+    def __repr__(self):
+        """Represent AtomArray as a string for debugging."""
+        atoms = ''
+        for i in range(0, self.array_length()):
+            if len(atoms) == 0:
+                atoms = self.get_atom(i).__repr__()
+            else:
+                atoms = atoms + ',\n' + self.get_atom(i).__repr__()
+        return f'array([{atoms}])'
```
padix-key, line 685: Could you put each atom on a separate line with indentation? Currently, the output is
```
array([Atom(...)
...
Atom(...)])
```

but I think
```
array([
    Atom(...)
    ...
    Atom(...)
])
```
would be better readable.

src/biotite/structure/atoms.py
```diff
@@ -909,7 +932,17 @@ def __init__(self, depth, length):
...
+
+    def __repr__(self):
+        """Represent AtomArray as a string for debugging."""
+        arrays = ''
+        for i in range(0, self.stack_depth()):
+            if len(arrays) == 0:
+                arrays = self.get_array(i).__repr__()
+            else:
+                arrays = arrays + ',\n' + self.get_array(i).__repr__()
+        return f'stack([{arrays}])'
```
padix-key, line 944: Similar to the comment above, i would like to request another layer of indentation and an empty line between each model to visibly separate them.

padix-key (changes requested):
I have a few minor points, which I highlighted in the comments below.

padix-key (approved):
Thank you. I will merge the PR now. The failing AppVeyor tests are an independent issue in the package setup.

PR #309, opened 2021-05-08
UniProt

MaxGreil:
Implementation of biotite interface to UniProt database.

This PR implements:

- uniprot.download
- uniprot.query

### Question:
- @padix-key , please have a look here https://www.uniprot.org/help/api%5Fretrieve%5Fentries:
I don't know which databases of UniProt should be accessible and which file formats should be downloadable 
(currently accessible databases are UniProtKB, UniRef and UniPark, file format is fasta)
It seems that available file formats for entries between the databases and even inside the same database can vary


padix-key:
In the best case, all of these database and formats should be supported. However, I think *UniProtKB* and FASTA + GFF are the most relevant ones. The identifiers seem to be unique for each database: For example all identifiers starting with `UPI` seem to correspond to *Uniparc* entries. Can you confirm this? If this is the case, it is not necessary for the user to set the database name, but it's sufficient to give the entry identifier.

> It seems that available file formats for entries between the databases and even inside the same database can vary

After the database name is inferred from the given identifier(s), the function might check whether the requested format is available for this database and raise a `ValueError` if it is not available.
If the available formats for different entries within the same database vary, I think requesting the file and raise a `RequestError` for an erroneous response is the only solution.

MaxGreil:
@padix-key I applied your suggested changes. Can you have another look?

src/biotite/database/uniprot/download.py
```diff
@@ -0,0 +1,190 @@
...
+                content = r.text
+                _assert_valid_file(r.status_code, id, format)
+            elif format == "rdf":
+                r = requests.get(fetch_url + db_name + "/" + id + ".rdf")
+                content = r.text
+                _assert_valid_file(r.status_code, id, format)
+            elif format == "tab":
+                r = requests.get(fetch_url + db_name + "/" + id + ".tab")
+                content = r.text
+                _assert_valid_file(r.status_code, id, format)
```
padix-key, line 148: This if-else-block contains quite a lot of repetitions. Alternatively, you could check for a valid file format at first and then handle the request for any format using the same code.

src/biotite/database/uniprot/download.py
```diff
@@ -0,0 +1,190 @@
...
+    """
+    if response_status_code == 400:
+        raise RequestError("Bad request. There is a problem with your input {:}.".format(uid + "." + format))
+    elif response_status_code == 404:
+        raise RequestError("Not found. The resource {:} you requested doesn't exist.".format(uid + "." + format))
+    elif response_status_code == 410:
+        raise RequestError("Gone. The resource {:} you requested was removed.".format(uid + "." + format))
+    elif response_status_code == 500:
+        raise RequestError(
+            "Internal server error. Most likely a temporary problem, but if the problem persists please contact us.")
```
padix-key, line 188: In my opinion *us* sounds like you refer to the Biotite developers. Maybe you could use *UniProt team* instead?

tests/database/test_uniprot.py
```diff
@@ -0,0 +1,33 @@
...
+import pytest
+import biotite.database.uniprot as uniprot
+import biotite.sequence.io.fasta as fasta
+from ..util import cannot_connect_to
+
+
+UNIPROT_URL = "https://www.uniprot.org/"
+
+
+pytest.mark.skipif(
```
padix-key, line 16: `skipif()` is a decorator, so you need to add an ``@`` before it to take effect. Furthermore, I think the two empty lines after the `skipif()` statement should be removed, so it gets clearer, that the decorator is associated to `test_fetch()`.

tests/database/test_uniprot.py
```diff
@@ -0,0 +1,33 @@
...
+    cannot_connect_to(UNIPROT_URL),
+    reason="UniProt is not available"
+)
+
+
+@pytest.mark.parametrize(
+    "as_file_like",
+    itertools.product([False, True])
+)
+def test_fetch(as_file_like):
```
padix-key, line 26: Could you also add test cases for an ID from *UniRef* and *UniParc*

src/biotite/database/uniprot/download.py
```diff
@@ -0,0 +1,190 @@
...
+    return "uniprot"
+
+
+def fetch(uids, format, target_path=None,
+          overwrite=False, verbose=False):
+    """
+    Download structure files (or sequence files) from the UniProt in
+    various formats.
+
+    Available databases are UniProtKB, UniRef and UniParc
```
padix-key, line 45: Missing period.

src/biotite/database/uniprot/download.py
```diff
@@ -0,0 +1,190 @@
...
+from os.path import isdir, isfile, join, getsize
+import os
+import io
+import requests
+from ..error import RequestError
+
+fetch_url = "https://www.uniprot.org/"
+
+
+def _get_database_name(uid):
```
padix-key, line 18: I think in *UniProt* the IDs are not called *UIDs* in contrast to the NCBI databases.

src/biotite/database/uniprot/download.py
```diff
@@ -0,0 +1,190 @@
...
+    Examples
+    --------
+
+    >>> import os.path
+    >>> file = fetch("P12345", "fasta", path_to_directory)
+    >>> print(os.path.basename(file))
+    P12345.fasta
+    >>> files = fetch(["P12345", "Q8K9I1"], "fasta", path_to_directory)
+    >>> print([os.path.basename(file) for file in files])
+    ['P12345.fasta', 'Q8K9I1.fasta']
```
padix-key, line 96: Could you also add the `biotite.database.uniprot` subpackage as test case to `tests/test_doctest.py::test_doctest()`? This way the doctest code output is tested by *pytest*.

padix-key (changes requested):
Thank you for the improvements. The remaining changes I would like to request are included in this code review.

*UniProt* can also handle queries (https://www.uniprot.org/help/api%5Fqueries) to search in the database. Do you also plan to add support for this API in this PR or is this PR only for the `fetch()` functionality?

MaxGreil:
@padix-key I applied your requested changes. I also implemented uniprot.query. Can you have another look?

src/biotite/database/uniprot/query.py
```diff
@@ -0,0 +1,201 @@
...
+__all__ = ["SimpleQuery", "CompositeQuery", "search"]
+
+import requests
+import abc
+
+_base_url = "https://www.uniprot.org/uniprot/"
+
+_search_url = ("?query={:}"
+               "&format=list"
+               "&limit={:}")
```
padix-key, line 16: This part of the URL can be described more clearly by the `params` parameter of `requests.get()`

tests/database/test_uniprot.py
```diff
@@ -0,0 +1,71 @@
...
+        )
+
+
+@pytest.mark.skipif(
+    cannot_connect_to(UNIPROT_URL),
+    reason="UniProt is not available"
+)
+def test_search_basic():
+    query = uniprot.SimpleQuery("accession", "P62988")
+    assert uniprot.search(query) == ['P62979', 'P0CG47', 'P62987', 'P0CG48', 'P62988']
```
padix-key, line 71: Could you also add a test for `CompositeQuery`?

src/biotite/database/uniprot/query.py
```diff
@@ -0,0 +1,201 @@
...
+    database might return no or malformed data to you.
+    In these cases the request should be retried.
+    When the issue occurs repeatedly, the error is probably in your
+    input.
+
+    Notes
+    -----
+    A list of available search fields with description can be found
+    `here <https://www.uniprot.org/help/query-fields>`_.
+    """
```
padix-key, line 195: I think an example code snippet would be helpful here, so the user can additionally see how combination of the `SimpleQuery` works.

padix-key (changes requested):
Thank you. I would like to request some changes in `query.py` and I found some typos in `download.py`. In general the code looks good to me :+1:.

MaxGreil:
@padix-key I applied your changes.

Regarding your question about database responses: 
based on the status codes for UniProt (https://www.uniprot.org/help/api_retrieve_entries), I think this happens in UniProt, too.

Can you please have another look?

padix-key:
In case of *NCBI Entrez* 'no or malformed data' refers to an *OK* status code, but still the returned file is malformed or empty. So no exception is raised on fetching the file, since the status code is *OK*, but an error occurs downstream, when the file is read.

MaxGreil:
@padix-key , thank you for the explanation.

Does this mean the 'Warnings' section from uniprot.download.fetch and uniprot.query.search can be removed?

Is the rest of this PR ready to merge?

padix-key:
> Does this mean the 'Warnings' section from uniprot.download.fetch and uniprot.query.search can be removed?

Yes, I think so.

> Is the rest of this PR ready to merge?

Your code looks good to me, so I will merge it after you remove the warning.

MaxGreil:
@padix-key , I removed the 'Warnings' section.

padix-key (approved):
Thank you! I will merge this PR now.

padix-key:
There is a failing test in the `database.rcsb` module due to outdated data of the PDB. I fixed the test in `biotite-dev biotite/master`. Could you merge it into the PR branch?

MaxGreil:
@padix-key , I merged the new master branch with uniprot

PR #310, opened 2021-05-10
Resolves #308

### LINKED ISSUE #308: io.load_structure filter template for atom_i
The `atom_i` parameter of `biotite.structure.io.xtc.XTCFile.read` allows to only read a subset of atoms into memory. If the trajectory is read via `io.load_structure(trj, template)`, this argument is not properly applied since the template is not filtered beforehand, leading to the following error:

Example:
```python
import biotite.structure.io as io
s = io.load_structure("nvt.xtc", "nvt.gro", atom_i=[1,2])
# ValueError: Template has 196252 atoms, but 2 coordinates are given
```

To work around this, the template must be filtered manually. I.e. this snippet works:
```python
import biotite.structure.io as io
s = io.load_structure("nvt.gro")
s = s[[1,2]]
s = io.load_structure("nvt.xtc", template=s, atom_i=[1,2])
```

This is a bit clumsy and also not intuitive to fix for a user that is not familiar with the underlying code of biotite. Maybe we can make it so that the atom_i filter is automatcally applied to the template if `atom_i.shape != template.shape` ? What do you think?

danijoo:
Here is a proposal for a way to filter templates based on `atom_i` in `io.load_structures`

### DIFF
--- a/src/biotite/structure/io/general.py
+++ b/src/biotite/structure/io/general.py
@@ -37,7 +37,8 @@ def load_structure(file_path, template=None, **kwargs):
         Additional parameters will be passed to either the
         :func:`get_structure()` or :func:`read()` method of the file
         object.
-        This does not affect files given via the `template` parameter.
+        This does not affect files given via the `template` parameter,
+        except for `atom_i`, which is applied to the template as well.
     
     Returns
     -------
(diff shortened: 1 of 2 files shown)


padix-key:
Thank you for the fix. Could you note in the docstring that the index is only applied to the template, if the number of atoms does not match the number of indices?

Issue #289, opened 2021-02-23
NetworkX for resolving conflicting pseudoknot cliques


padix-key:
With *NetworkX* becoming a *Biotite* dependency in the next update (#283), we could think about replacing the functions for resolving conflicting regions in `biotite.structure.pseudoknots()` with graph based solutions:

- `_find_non_conflicting_regions()` -> `networkx.isolates()`
- `_conflict_cliques()` -> `networkx.connected_components()`

This could make the code cleaner and even increase the performance (although this need to be tested).

@tomtomhdx What is your opinion about this?

tomtomhdx:
This sounds like an excellent idea!

I will make the adjustments and check whether the performance increases.

PR #320, opened 2021-05-27
Use NetworkX for resolving conflicting pseudoknot cliques

tomtomhdx:
As suggested in #289, this PR uses the new *NetworkX* dependency in order to resolve conflicting pseudoknots cliques.

Specifically the functions ``_find_non_conflicting_regions()`` and ``_conflict_cliques()`` were replaced by graph based solutions. While this change does increase code readability, there also seems to be a slight performance enhancement:

The run time of the corresponding test functions was 133s prior to the change. This PR reduces the running time to 109s.


padix-key:
Thank you for the improvement. From my side the code looks good, so I will merge the PR as soon as the tests have passed.

PR #324, opened 2021-06-09
Project 'Support for sequence profiles'

MaxGreil:
This PR aims to implement project 'Support for sequence profiles'

- Implementation of new new class SequenceProfile
- Implementation of tests

@padix-key can you take a look?


src/biotite/sequence/profile.py
```diff
@@ -0,0 +1,165 @@
...
+    position_gaps : ndarray, dtype=int, shape=n
+        Array which indicates the number of gaps at each position.
+    alphabet : Alphabet, length=k
+        Alphabet of sequences of sequence profile
+    """
+
+    def __init__(self, position_frequency_matrix, position_gaps, alphabet):
+        self.position_frequency_matrix = position_frequency_matrix
+        self.position_gaps = position_gaps
+        self.alphabet = alphabet
```
padix-key, line 55: I suggest that the public access is handled via `@property`. For example I think the user should not be able to change the alphabet, because then the size of the frequency matrix does not fit anymore. Instead I think `position_frequency_matrix` and `position_frequency_matrix` should check whether the shape is still the same, when these attributes are set.

Furthermore, the variable names `position_frequency_matrix` and `position_frequency_matrix` are quite long in my opinion. May I suggest `symbols` and `gaps` instead?

src/biotite/sequence/profile.py
```diff
@@ -0,0 +1,165 @@
...
+        if not np.array_equal(self.position_gaps, item.position_gaps):
+            return False
+        if not self.alphabet == item.alphabet:
+            return False
+        return True
+
+    @staticmethod
+    def from_alignment(alignment):
+        """
+        Get an object of :class:`SequenceProfile` from an object of :class:`Alignment`
```
padix-key, line 76: Could you elaborate in the docstring how an alignment ins converted into a profile?

src/biotite/sequence/profile.py
```diff
@@ -0,0 +1,165 @@
...
+        """
+        sequences = alignment.get_gapped_sequences()
```
padix-key, line 78: Using `get_gapped_sequences()` in this place is quite inefficient: The trace is converted to symbol codes, the symbol codes to symbols and the symbols into a string, and vice versa.

Probably `sequence.align.get_codes()` (https://www.biotite-python.org/apidoc/biotite.sequence.align.get_codes.html) is more efficient in this place. Then you can use `np.bincount()`for every column. You could handle `-1` values by replacing them with `len(alphabet)`

src/biotite/sequence/profile.py
```diff
@@ -0,0 +1,165 @@
...
+        alphabet = alignment.sequences[0].get_alphabet().get_symbols()
```
padix-key, line 79: Could you check if all sequences in the alignment use the same alphabet?

src/biotite/sequence/profile.py
```diff
@@ -0,0 +1,165 @@
...
+        elif set(self.alphabet.get_symbols()) == set(_NUC_RNA_ALPH.get_symbols()):
+            return self._rna_to_consensus()
+        elif set(self.alphabet.get_symbols()) == set(_PROT_ALPH.get_symbols()):
+            return self._prot_to_consensus()
+        return self._custom_to_consensus()
+
+    def _codes_to_iupac(self, frequency, codes):
+        if np.sum(frequency) == 0:
+            return "-"
+        position_max = np.where(frequency == np.amax(frequency))[0].tolist()
```
padix-key, line 111: The current implementation runs this for every position in the profile. However `np.where()` and `np.amax()` are also able to handle multidimensional arrays. So maybe it is possible to run this once for the entire profile to increase the efficiency.

src/biotite/sequence/profile.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def to_consensus(self):
+        """
+        Get the consensus sequence for :class:`SequenceProfile`
+
+        Returns
+        -------
+        consensus: str
+            The calculated consensus sequence
+        """
```
padix-key, line 98: I think it is more consistent with the rest of the package if this function returns the consensus as `Sequence` object, either as `NucleotideSequence`, `ProteinSequence` or `GeneralSequence` (for custom alphabets) object. Maybe a parameter like `as_general` would be helpful, if the user wants a `GeneralSequence` in any case.

tests/sequence/test_profile.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def test_to_consensus_nuc():
+    position_frequency_matrix = np.array([[0, 1, 0, 0], [0, 0, 1, 0], [0, 0, 0, 2], [0, 2, 0, 0],
+                                          [2, 0, 0, 0], [0, 0, 0, 2], [0, 0, 1, 0], [0, 1, 0, 0]])
+    position_gaps = np.array([1, 1, 0, 0, 0, 0, 1, 1])
+    alphabet = seq.Alphabet(["A", "C", "G", "T"])
+    profile = prof.SequenceProfile(position_frequency_matrix, position_gaps, alphabet)
+
+    assert "CGTCATGC" == profile.to_consensus()
```
padix-key, line 37: For completeness this function should also test the case where the resulting sequence contains an ambiguous symbol.

padix-key (changes requested):
Hi, thank you for the PR. Here is my review. I especially like the idea that the consensus sequence contains unambiguous symbols, if the input sequences have two different symbols at the same frequency.

src/biotite/sequence/profile.py
MaxGreil: Hi @padix-key , I chose this implementation because I need to get multiple maximum values for each row of ndarray symbols. I did not find a solution to to this once for a multidimensional array. Do you have an idea how to do this?

MaxGreil:
Hi @padix-key , I applied your requested changes. Can you have another look?

src/biotite/sequence/profile.py
padix-key: For `np.where()` you are probably right, I also have currently no direct solution to solve this in a single function call. However, you could at least obtain the maximum value in each column with `np.max(self._symbols, axis=1)`

src/biotite/sequence/profile.py
```diff
@@ -0,0 +1,211 @@
...
+            return NucleotideSequence(self._rna_to_consensus())
+        elif set(self.alphabet.get_symbols()) == set(_PROT_ALPH.get_symbols()):
+            return ProteinSequence(self._prot_to_consensus())
+        return GeneralSequence(self.alphabet, self._custom_to_consensus())
+
+    def _codes_to_iupac(self, frequency, codes):
+        if np.sum(frequency) == 0:
+            return "-"
+        position_max = np.where(frequency == np.amax(frequency))[0].tolist()
+        key = "".join(["".join(self.alphabet.get_symbols()[i]) for i in position_max])
```
padix-key, line 158: Calling `self.alphabet.get_symbols()` for each column introduces some overhead, since the list of symbols is copied everytime. it is probably more efficient to use `Alphabet.decode()` instead or, even better, to directly use symbols codes and skip using characters entirely.

padix-key (changes requested):
Hi, thank you for the changes. I left some further comments. Although I think most of the code will already work as intended, there are still some issues regarding the performance in my opinion.

MaxGreil:
Hi @padix-key , I applied the changes. Can you have another look?

src/biotite/sequence/profile.py
```diff
@@ -0,0 +1,244 @@
...
+            return NucleotideSequence(self._dna_to_consensus())
+        elif self.alphabet == _NUC_RNA_ALPH:
+            return NucleotideSequence(self._rna_to_consensus())
+        elif self.alphabet == _PROT_ALPH:
+            return self._prot_to_consensus()
+        return self._general_to_consensus()
+
+    def _dna_to_consensus(self):
+        codes = {
+            '0': 'A', '1': 'C', '2': 'G', '3': 'T',
```
padix-key, line 202: See above

src/biotite/sequence/profile.py
```diff
@@ -0,0 +1,244 @@
...
+        }
+        consensus = ""
+        maxes = np.max(self.symbols, axis=1)
+        for i in range(len(self.symbols)):
+            consensus += _codes_to_iupac(self.symbols[i, :], codes, maxes[i])
+        return consensus
+
+    def _rna_to_consensus(self):
+        codes = {
+            '0': 'A', '1': 'C', '2': 'G', '3': 'U',
```
padix-key, line 215: See above

src/biotite/sequence/profile.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+def _codes_to_iupac(frequency, codes, maxes):
+    """
+    Returns IUPAC code for a row of 'symbols' with none, one or
+    multiple maximum positions.
+    """
+    if np.sum(frequency) == 0:
+        return "-"
+    key = ''.join(np.where(frequency == maxes)[0].astype(str))
```
padix-key, line 44: Can you use a tuple of integers instead of a string as dictionary key? This would remove the time-costly integer-to-string conversion. Furthermore, I think that a tuple of integers makes it more clear that they key is based on sequence codes.

src/biotite/sequence/profile.py
```diff
@@ -0,0 +1,244 @@
...
+                [seq.alphabet for seq in alignment.sequences]
+            )
+        else:
+            for alph in (seq.alphabet for seq in alignment.sequences):
+                if not alphabet.extends(alph):
+                    raise ValueError(
+                        "The given alphabet is incompatible with a least one "
+                        "alphabet of the given sequences"
+                    )
+            alphabet = alphabet
```
padix-key, line 162: I think, this is basically a *no-op* you can omit.

padix-key (commented):
Thanks for the additional work. As a final change, i would suggest using tuples as dictionary keys for getting the consensus sequence. After this I would merge this PR, if you have no objections.

MaxGreil:
Hi @padix-key , I changed to tuples. I have no objects to merge this PR. I will then add the missing features for project sequence profiles in another PR.

padix-key (approved):
OK, thank you for your contribution! :+1:

PR #335, opened 2021-06-29
Use sphinxcontrib-bibtex for citation management

padix-key:
This PR introduces the usage of `sphinxcontrib-bibtex` for citation management in the entire project. Although this adds a new dependency for building the documentation and requires contributors to use a *BibTeX* file, this change has some significant advantages:

- Reduction of potentially malformed citations
- Enforced consistent citation style
- Easy way to globally change the citation style
- Adding a DOI to each of the existing citations


padix-key:
@tomtomhdx @danijoo If I remember correctly, you already used citations in dcostrings. Do you have an opinion about adding citation information to a BibTeX file compared to manual citations in a docstring?

danijoo:
I like bibtex because they can be more easily exported and are consistent across the repo.

tomtomhdx:
I agree with @danijoo.

Also I would imagine it is more comfortable to enter citations in the BibTeX format than having to manually match the right citation format in the docstring.

padix-key:
Thank you for your feedback. In this case I will merge this PR.

PR #341, opened 2021-07-15
Issue 337

### LINKED ISSUE #337: SequenceProfile does not check input parameters
Currently, the constructor of `SequenceProfile` does not check the shapes of the parameters given to the constructor. This may lead to unexpected errors later: For example, the alphabet is not able to decode a symbol code in `to_consensus()` if the alphabet is shorter than the number of columns in the symbol symbol frequency table.

MaxGreil:
Fixes #337

Implements additional checks for profile constructor.

@padix-key , can you have a look?

### DIFF
--- a/src/biotite/sequence/profile.py
+++ b/src/biotite/sequence/profile.py
@@ -87,6 +87,20 @@ class SequenceProfile(object):
         self._gaps = gaps
         self._alphabet = alphabet
 
+        if len(alphabet) != self.symbols.shape[1]:
+            raise ValueError(
+                f"The given alphabet is shorter ({len(self.alphabet)}) "
+                f"than the number of columns ({self.symbols.shape[1]}) in "
+                f"the 'symbols' frequency table."
+            )
+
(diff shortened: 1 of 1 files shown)


src/biotite/sequence/profile.py
```diff
@@ -87,6 +87,20 @@ def __init__(self, symbols, gaps, alphabet):
         self._gaps = gaps
         self._alphabet = alphabet
 
+        if len(alphabet) != self.symbols.shape[1]:
+            raise ValueError(
+                f"The given alphabet is shorter ({len(self.alphabet)}) "
```
padix-key, line 92: It could also be the case, that the alphabet is longer than the number of columns.

padix-key (changes requested):
Thanks @MaxGreil . I think there is a potentially misleading formulation in the error message.

MaxGreil:
Hi @padix-key , I revised the error text. Can you have another look?

padix-key (approved):
Looks good to me, thank you!

PR #342, opened 2021-07-19
Issue 338

### LINKED ISSUE #338: 'to_consensus()' raises unexpected exception
If at least one column of in the symbol frequency matrix  of `SequenceProfile` is empty and `to_consensus()` is called, an unexpected exception is raised:

```python
import numpy as np
import biotite.sequence as seq

SEQ_LENGTH = 10

alphabet = seq.NucleotideSequence.alphabet_unamb
profile = seq.SequenceProfile(
    np.zeros((SEQ_LENGTH, len(alphabet)), dtype=int),
    np.zeros(SEQ_LENGTH, dtype=int),
    alphabet
)

print(profile.to_consensus())
```

```
biotite.sequence.AlphabetError: Symbol '-' is not in the alphabet
```

The reason for this error is here:

https://github.com/biotite-dev/biotite/blob/05e8111e69ed43f51783b683b1701f92e7017b09/src/biotite/sequence/profile.py#L42-L43

A column without symbols results in `-1`, which is mapped to `'-'`, which is not a valid symbol in the nucleotide and amino acid alphabet.

This exception should not happen in most use cases, since an empty column makes no sense in context of an alignment.
I think a solution to this would be either returning the ambiguous symbol (`X` or `N`) in this case or raising an exception that indicates the empty column.

MaxGreil:
Fixes #338 

### DIFF
--- a/src/biotite/sequence/profile.py
+++ b/src/biotite/sequence/profile.py
@@ -34,13 +34,17 @@ def _determine_common_alphabet(alphabets):
     return common_alphabet
 
 
-def _codes_to_iupac(frequency, codes, maxes):
+def _codes_to_iupac(frequency, codes, maxes, row):
     """
     Returns IUPAC code for a row of 'symbols' with none, one or
     multiple maximum positions.
     """
     if np.sum(frequency) == 0:
-        return codes[(-1,)]
+        raise ValueError(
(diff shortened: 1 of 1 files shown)


padix-key:
Thank you for the fix. :+1:

PR #346, opened 2021-08-15
Implement PPM and PWM

MaxGreil:
This PR is work in progress.

First try to add a position probability matrix and a position weight matrix to class SequenceProfile.

Changed files:
- src/biotite/sequence/profile.py
- tests/sequence/test_profile.py


padix-key:
Thank you for the PR. I am currently on vacation until end of the week, but I will have a look at it next week.

MaxGreil:
@padix-key I applied your requested changes. Can you have a look?

src/biotite/sequence/profile.py
```diff
@@ -280,3 +291,88 @@ def _general_to_consensus(self):
         consensus = GeneralSequence(self.alphabet)
         consensus.code = np.argmax(self.symbols, axis=1)
         return consensus
+
+    def probability_matrix(self, pseudocount=0):
+        """
+        Calculate the position probability matrix (ppm) based on 'symbols'
```
padix-key, line 297: Since `ppm` is an abbreviation, I would write it in uppercase.

src/biotite/sequence/profile.py
```diff
@@ -280,3 +291,88 @@ def _general_to_consensus(self):
...
+        and the given pseudocount.
+        This new matrix has the same shape as 'symbols'.
+
+        Parameters
+        ----------
+        pseudocount: int
```
padix-key, line 303: Since this parameter is optional, it should be written as `pseudocount: int, optional`

src/biotite/sequence/profile.py
```diff
@@ -280,3 +291,88 @@ def _general_to_consensus(self):
...
+            Amount added to the number of observed cases in order to change
+            the expected probability.
+            (Default: 0)
+
+        Returns
+        -------
+        ppm: ndarray, dtype=float, shape=(n,k)
+            Calculated the position probability matrix.
```
padix-key, line 311: Did you mean to write `The calculated position probability matrix.`, or am I wrong here?

src/biotite/sequence/profile.py
```diff
@@ -280,3 +291,88 @@ def _general_to_consensus(self):
...
+        """
+        if pseudocount < 0:
+            raise ValueError(
+                f"Pseudocount can not be smaller than zero."
+            )
+        return (self.symbols + pseudocount / len(self.symbols)) / \
```
padix-key, line 317: I think the correct term is `(self.symbols + pseudocount / len(self.alphabet))` or `(self.symbols + pseudocount / self.symbols.shape[1])`. Can you confirm that?

src/biotite/sequence/profile.py
```diff
@@ -280,3 +291,88 @@ def _general_to_consensus(self):
...
+        pseudocount: int
+            Amount added to the number of observed cases in order to change
+            the expected probability.
+            (Default: 0)
+        background_frequencies: float
+            (Default: 1/len(alphabet))
+        Returns
+        -------
+        pwm: ndarray, dtype=float, shape=(n,k)
+            Calculated the position weight matrix.
```
padix-key, line 338: See above

src/biotite/sequence/profile.py
```diff
@@ -280,3 +291,88 @@ def _general_to_consensus(self):
...
+            )
+        if not matrix.shape == self.symbols.shape:
+            raise ValueError(
+                f"Matrix {matrix.shape} must be of same shape "
+                f"as 'symbols' {self.symbols.shape}"
+            )
+        probability = 1
+        codes = sequence.code
+        for i in range(len(codes)):
+            probability = probability * matrix[i, codes[i]]
```
padix-key, line 377: I think this loop can be vectorized with `numpy.prod(matrix[np.arange(len(sequence)), sequence.code])`

src/biotite/sequence/profile.py
```diff
@@ -280,3 +291,88 @@ def _general_to_consensus(self):
...
+        """
+        if background_frequencies is None:
+            background_frequencies = 1 / len(self.alphabet)
+        ppm = self.probability_matrix(pseudocount)
+        return np.log2(ppm / background_frequencies)
+
+    def sequence_probability_from_matrix(self, sequence, matrix):
+        """
+        Calculate probability of a sequence based on either the
+        position probability matrix (ppm) or the position weight matrix (pwm).
```
padix-key, line 348: Could you reference the mathematical background for the calculation the sequence probability from a position weight matrix? In my intuition the product of values from the position weight matrix does not represent a probability, since each value represents the logarithm of a probability, but I may be wrong.

padix-key (changes requested):
Thanks for the work, here is my review. If you like, you could also add the underlying formulas to the docstring, if you are versed in LaTeX. Otherwise, I would volunteer to add them, if you agree.

MaxGreil:
@padix-key I applied the changes.
I don't know Latex. Could you add the underlying formulas?
And you were correct about the sequence probability from a position weight matrix. I looked it up and you can only calculate a sequence 'score' from a PWM.

padix-key:
Since the `matrix` parameter in `sequence_probability()` and `sequence_score()` is unambiguous now, it can be calculated from the `SequenceProfile` object itself. Hence, I think this parameter can be removed from both functions.

MaxGreil:
@padix-key I implemented the latest changes.

padix-key (approved):
The changes look good to me :+1:. I will add the respective formulas in LaTeX to the docstrings in the next few days, before I merge the PR.

MaxGreil:
@padix-key your additional changes look good to me. Can this PR now be merged?

padix-key:
I just changed the assertion expression in the tests to shorter ones using `np.allclose()` and `pytest.approx()` and wait for the tests to finish.

PR #364, opened 2021-11-27
fix #360

### LINKED ISSUE #360: Spacing in Alignments
In alignments of all kind it would help very much if one could insert spacings. This gives a better overview and helps with finding e.g. specifc amino acid residues.
For protein sequences for example a spacing after every 10 amino acids is appropiate. For nucleic acid sequences a spacing after 9 bases helps to keep it in parallel with the open reading frame.

alex123012:
* add desired spacing between elemnets in alignments (function `sequence.graphics.plot_alignment` and its derived functions). for e.g.:
```python
# Code source: Patrick Kunzmann
# License: BSD 3 clause

import biotite.sequence as seq
import biotite.sequence.align as align
import biotite.sequence.io.fasta as fasta
import biotite.database.entrez as entrez
import biotite.sequence.graphics as graphics

# Download and parse protein sequences of avidin and streptavidin
fasta_file = fasta.FastaFile.read(entrez.fetch_single_file(
    ["CAC34569", "ACL82594"], None, "protein", "fasta"
))
for name, sequence in fasta_file.items():
    if "CAC34569" in name:
        avidin_seq = seq.ProteinSequence(sequence)
    elif "ACL82594" in name:
        streptavidin_seq = seq.ProteinSequence(sequence)

# Get BLOSUM62 matrix
matrix = align.SubstitutionMatrix.std_protein_matrix()
# Perform pairwise sequence alignment with affine gap penalty
# Terminal gaps are not penalized
alignments = align.align_optimal(avidin_seq, streptavidin_seq, matrix,
                                 gap_penalty=(-10, -1), terminal_penalty=False)

# Draw first and only alignment
# The color intensity indicates the similiarity
fig = plt.figure(figsize=(8.0, 2.5))
ax = fig.add_subplot(111)
graphics.plot_alignment_similarity_based(
    ax, alignments[0], matrix=matrix, labels=["Avidin", "Streptavidin"],
    show_numbers=True, show_line_position=True, symbol_spacing=5, symbols_per_line=50
)
fig.tight_layout()

plt.show()
```
image

*  Some code style refactoring in `src/biotite/sequence/graphics/alignment.py` file

### DIFF
--- a/src/biotite/sequence/graphics/alignment.py
+++ b/src/biotite/sequence/graphics/alignment.py
@@ -11,8 +11,7 @@ __all__ = ["SymbolPlotter", "LetterPlotter", "LetterSimilarityPlotter",
 
 import abc
 import numpy as np
-from ...visualize import set_font_size_in_coord, colors
-from ..seqtypes import ProteinSequence
+from ...visualize import colors
 from .colorschemes import get_color_scheme
 
 
@@ -22,7 +21,7 @@ class SymbolPlotter(metaclass=abc.ABCMeta):
     an alignment are drawn onto an :class:`Axes` object.
 
     Subclasses must override the :func:`plot_symbol()` method.
-    
+
     Parameters
     ----------
     axes : Axes
@@ -31,7 +30,7 @@ class SymbolPlotter(metaclass=abc.ABCMeta):
 
     def __init__(self, axes):
         self._axes = axes
-    
+
     @property
     def axes(self):
         return self._axes
@@ -66,7 +65,7 @@ class LetterPlotter(SymbolPlotter, metaclass=abc.ABCMeta):
     as colored character, if `color_symbols` is set to true.
 
     Subclasses must override the :class:`get_color()` method.
-    
+
     Parameters
     ----------
     axes : Axes
@@ -82,7 +81,7 @@ class LetterPlotter(SymbolPlotter, metaclass=abc.ABCMeta):
         :class:`matplotlib.Text` instance of each symbol.
     """
 
-    def __init__(self, axes, color_symbols=False, 
+    def __init__(self, axes, color_symbols=False,
                  font_size=None, font_param=None):
         super().__init__(axes)
         self._color_symbols = color_symbols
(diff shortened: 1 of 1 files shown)


alex123012:
check symbol_spacing flag in plot_alignment function

src/biotite/sequence/graphics/alignment.py
```diff
@@ -445,6 +445,11 @@ def plot_alignment(axes, alignment, symbol_plotter, symbols_per_line=50,
     if seq_len % symbols_per_line != 0:
         line_count += 1
 
+    if symbol_spacing:
+        spacing_ratio = symbols_per_line / symbol_spacing
+        symbols_to_print = round(spacing_ratio) + symbols_per_line - 1
```
padix-key, line 450: I do not quite understand the aim of calculating `symbols_to_print` from `symbols_per_line`. I think a comment line would be helpful here.

src/biotite/sequence/graphics/alignment.py
```diff
@@ -408,7 +405,10 @@ def plot_alignment(axes, alignment, symbol_plotter, symbols_per_line=50,
     spacing : float, optional
         The spacing between the alignment lines. 1.0 means that the size
         is equal to the size of a symbol box.
-    
+    symbol_spacing : int, optional
+        А space is placed between each number of elements desired
+        by variable (github issue #360)
```
padix-key, line 410: GitHub issues are not referenced in docstrings in Biotite, at least no in the `Parameters` section.

```suggestion
        by variable.
```

src/biotite/sequence/graphics/alignment.py
```diff
@@ -629,7 +639,10 @@ def plot_alignment_similarity_based(axes, alignment, symbols_per_line=50,
     symbol_param : dict
         Additional parameters that is given to the
         :class:`matplotlib.Text` instance of each symbol.
-    
+    symbol_spacing : int, optional
+        А space is placed between each number of elements desired
+        by variable (github issue #360)
```
padix-key, line 644:
```suggestion
        by variable.
```

src/biotite/sequence/graphics/alignment.py
```diff
@@ -760,7 +774,10 @@ def plot_alignment_type_based(axes, alignment, symbols_per_line=50,
...
```
padix-key, line 779:
```suggestion
        by variable.
```

padix-key (changes requested):
Thank you very much for implementing the feature as well as cleaning up the code in the module! I tested the code on some examples from the library and it worked very well in most cases. However, I found any issue in the edge case, where `symbols_per_line` is not a multiple of `symbol_spacing`:

Figure_1

Starting from the second line one alignment column is missing at the right end of the line, which also leads to a wrong sequence position number of the rightmost column.

Instead of solving this issue, I think it would be also OK to forbid this edge case, by raising an exception, if  `symbols_per_line` is not a multiple of `symbol_spacing`. I leave this decision up to you.

alex123012 (commented):
Resolve all your comments. Excuse me, because of time, can't fully resolve issue with not multiple symbols_per_line by symbol_spacing. Maybe in the future I'll fix it.

padix-key (approved):
The PR looks good to me. You may also add yourself to the `CONTRIB.rst`, if you like.

padix-key:
I already approved the PR. The failing test is caused by an unrelated temporary error in the `database` subpackage. But before I merge, you can add yourself to the list of contributors if you like.

alex123012:
> I already approved the PR. The failing test is caused by an unrelated temporary error in the `database` subpackage. But before I merge, you can add yourself to the list of contributors if you like.

Oh, sry, didn't look on tests output, thanks for clarification.

PR #366, opened 2021-12-28
Update contribute.rst

entropybit:
Very minor typo, just saw this and thought I should correct it.


padix-key:
Thank you for the fix

PR #374, opened 2022-02-23
foldx Buildmodel mutation interfact - convenience interface

Mojusko:
Not sure this is interesting to you.


padix-key:
Hi, thank you for adding the interface. Initially, I have two requests: 1. Could you give a short description in the PR, what the interfaced software does? 2. Via a quick search I found out that this is a proprietary program. This is not a problem in general, but renders automatic testing in the CI pipeline difficult. Is there an open source version of the software?

Mojusko:
1. It introduces a mutation to a protein a generates a structure for it. Its part of larger suite which prepares mutated proteins for MD simulations 2. Indeed its proprietary, but so is vina which also has an interface. I would like to know open-source version of this as well. A similar functionality is contained in proprietary version of pymol. I was not sure if this is something interesting to this community. Anyway I like your package. Let me know if I should finish this.

padix-key:
Basically every interface for publicly available and well documented software that works on sequence or structure data is interesting for `biotite.application`. I will have a look on your PR and prepare a review.

Regarding your questions on open-source versions of AutoDock Vina and PyMOL:
To my knowledge Vina is exclusively open source (https://vina.scripps.edu/license/), while for PyMOL only a basic version (without officially built binaries) is open-source (https://github.com/schrodinger/pymol-open-source/blob/master/LICENSE).

src/biotite/application/foldx/app.py
```diff
@@ -0,0 +1,190 @@
+# This source code is part of the Biotite package and is distributed
+# under the 3-Clause BSD License. Please see 'LICENSE.rst' for further
+# information.
+
+__name__ = "biotite.application.foldX"
```
padix-key, line 5:
```suggestion
__name__ = "biotite.application.foldx"
```

src/biotite/application/foldx/app.py
```diff
@@ -0,0 +1,190 @@
...
+import os 
+from os import chdir, getcwd, remove
+from ..localapp import LocalApp, cleanup_tempfile
+from ..application import AppState, requires_state
+from ...structure.io.pdbqt import PDBQTFile
+from ...structure.io.pdb import PDBFile
+from ...structure.io.pdbx import PDBxFile
+from ...structure.residues import get_residue_starts_for, get_residue_masks
+from ...structure.bonds import find_connected
+from ...structure.error import BadStructureError
```
padix-key, line 21: A lot of these imports are actually unused. Please remove the unused imports, when you finished your work on the PR.

src/biotite/application/foldx/app.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+class FoldXApp(LocalApp):
```
padix-key, line 24: As I read so far, FoldX has multiple subcommands, and the one used here is called `BuildModel`. Hence, i propose the following:
```suggestion
class BuildModelApp(LocalApp):
```
This is consistent with e.g. the `viennarna` subpackage, where the classes are called `RNAfoldApp` and `RNAplotApp`.

src/biotite/application/foldx/app.py
```diff
@@ -0,0 +1,190 @@
...
+
+    @requires_state(AppState.JOINED)
+    def get_mutant(self):
+        """
+        Get the mutant protein structure
+
+        Returns
+        -------
+        ligand : AtomArrayStack
+            The mutated protein
```
padix-key, line 167:
```suggestion
            The mutated protein.
```

src/biotite/application/foldx/app.py
```diff
@@ -0,0 +1,190 @@
...
+        
+        """
+        return self.new_mutant
```
padix-key, line 170:
```suggestion
        return self._new_mutant
```

tests/application/test_foldx.py
```diff
@@ -0,0 +1,40 @@
...
+    structure = structure[structure.chain_id == "B"]
+    receptor = structure[struc.filter_amino_acids(structure)]
+    mutation = "A13A"
+
+    app = FoldXApp(receptor, mutation, subunit = 'B')
+    app.start()
+    app.join()
+    arr = app.get_mutant()
+    file = PDBFile()
+    assert (file is not None)
```
padix-key, line 40: I think this line is an error: A newly created `PDBFile` is never `None`.

tests/application/test_foldx.py
```diff
@@ -0,0 +1,40 @@
...
+        mmtf_file, model=1, extra_fields=["charge"], include_bonds=True
+    )
+    structure = structure[structure.chain_id == "B"]
+    receptor = structure[struc.filter_amino_acids(structure)]
+    mutation = "A13A"
+
+    app = FoldXApp(receptor, mutation, subunit = 'B')
+    app.start()
+    app.join()
+    arr = app.get_mutant()
```
padix-key, line 38: An actual check of the mutated output structure would be good here. Even if the actually output atom positions cannot be easily tested, I think it would be sufficient to check if the original atoms are still there, except the mutation position

tests/application/test_foldx.py
```diff
@@ -0,0 +1,40 @@
+# This source code is part of the Biotite package and is distributed
+# under the 3-Clause BSD License. Please see 'LICENSE.rst' for further
+# information.
+
+from os.path import join
+import numpy as np
+import pytest
+import biotite.structure as struc
+import biotite.structure.info as info
```
padix-key, line 9:
```suggestion
```

tests/application/test_foldx.py
```diff
@@ -0,0 +1,40 @@
+# This source code is part of the Biotite package and is distributed
+# under the 3-Clause BSD License. Please see 'LICENSE.rst' for further
+# information.
+
+from os.path import join
+import numpy as np
```
padix-key, line 6:
```suggestion
```

padix-key (changes requested):
Hi, I reviewed your FoldX interface and I think it is written well in general. However, I have two more major requests:

1. I think the docstrings are a bit sparse, especially regarding the purpose of the application and the description of the parameters. Since in my opinion the functionalities of the software are rather complex, I think a few more words would be helpful.

2. The `rotabase.txt` is packed into Biotite. While this is convenient to the user, I assume this is not allowed by the commercial FoldX license. If I am correct, this unfortunately means that the rotabase needs to be provided as additional input by the user.

I need to emphasize, that I am new to the FoldX software. Hence some of my comments may be incorrect.

PR #375, opened 2022-03-18
Added orient_principal_components

claudejrogers:
The function translates atoms to the origin and rotates the molecule
such that the variance decreases from x to z. Included a basic test.

### DIFF
--- a/src/biotite/structure/transform.py
+++ b/src/biotite/structure/transform.py
@@ -273,15 +273,15 @@ def orient_principal_components(atoms, order=None):
     of variance.
 
     >>> print("original variance =", atom_array.coord.var(axis=0))
-    original variance = [26.517471 20.008837  9.324988]
+    original variance = [26.517 20.009  9.325]
     >>> moved = orient_principal_components(atom_array)
     >>> print("moved variance =", moved.coord.var(axis=0))
-    moved variance = [28.906227 18.495403  8.449674]
+    moved variance = [28.906 18.495  8.450]
     >>> # Note the increase in variance along the x-axis
     >>> # Specifying the order keyword changes the orientation
     >>> moved_z = orient_principal_components(atom_array, order=(2, 1, 0))
     >>> print("moved (zyx) variance =", moved_z.coord.var(axis=0))
-    moved (zyx) variance = [ 8.449674 18.495403 28.906227]
+    moved (zyx) variance = [ 8.450 18.495 28.906]
     """
     # Check user input
     coords = coord(atoms)
--- a/src/biotite/structure/transform.py
+++ b/src/biotite/structure/transform.py
@@ -238,12 +238,12 @@ def orient_principal_components(atoms, order=None):
     """
     Translate and rotate the atoms to be centered at the origin with
     the principal axes aligned to the Cartesian axes, as specified by
-    the order keyword. By default, x, y, z.
+    the `order` parameter. By default, x, y, z.
 
     By default, the resulting coordinates have the highest variance in
-    the x-axis and the lowest variance on the z-axis. Setting the order
-    keyword argument will change the direction of the highest variance.
-    For example, order=(2, 1, 0) results in highest valianxe along the
+    the x-axis and the lowest variance on the z-axis. Setting the `order`
+    parameter will change the direction of the highest variance.
+    For example, ``order=(2, 1, 0)`` results in highest variance along the
     z-axis and lowest along the x-axis.
 
     Parameters
@@ -252,9 +252,9 @@ def orient_principal_components(atoms, order=None):
         The atoms of which the coordinates are transformed.
         The coordinates can be directly provided as :class:`ndarray`.
     order : array-like, length=3
-        The order of decreasing variance. Setting order to (2, 0, 1)
+        The order of decreasing variance. Setting `order` to ``(2, 0, 1)``
         results in highest variance along the y-axis and lowest along
-        the x-axis. Default = (0, 1, 2)
+        the x-axis. Default = ``(0, 1, 2)``.
 
     Returns
     -------
--- a/src/biotite/structure/transform.py
+++ b/src/biotite/structure/transform.py
@@ -8,7 +8,7 @@ that can be applied on structures.
 """
 
 __name__ = "biotite.structure"
-__author__ = "Patrick Kunzmann"
+__author__ = "Patrick Kunzmann", "Claude J. Rogers"
 __all__ = ["translate", "rotate", "rotate_centered", "rotate_about_axis",
            "orient_principal_components", "align_vectors"]
 
@@ -234,16 +234,27 @@ def rotate_about_axis(atoms, axis, angle, support=None):
     return _put_back(atoms, positions)
 
 
-def orient_principal_components(atoms):
+def orient_principal_components(atoms, order=None):
     """
-    Translate and rotate the atoms to be centered at the origin with the
-    principal axes aligned to x, y, and z.
+    Translate and rotate the atoms to be centered at the origin with
+    the principal axes aligned to the Cartesian axes, as specified by
+    the order keyword. By default, x, y, z.
+
+    By default, the resulting coordinates have the highest variance in
+    the x-axis and the lowest variance on the z-axis. Setting the order
+    keyword argument will change the direction of the highest variance.
+    For example, order=(2, 1, 0) results in highest valianxe along the
+    z-axis and lowest along the x-axis.
 
     Parameters
     ----------
     atoms : AtomArray or ndarray, shape=(n,3)
         The atoms of which the coordinates are transformed.
         The coordinates can be directly provided as :class:`ndarray`.
+    order : array-like, length=3
+        The order of decreasing variance. Setting order to (2, 0, 1)
+        results in highest variance along the y-axis and lowest along
+        the x-axis. Default = (0, 1, 2)
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


src/biotite/structure/transform.py
```diff
@@ -234,6 +234,78 @@ def rotate_about_axis(atoms, axis, angle, support=None):
     return _put_back(atoms, positions)
 
 
+def orient_principal_components(atoms):
```
padix-key, line 237: Would it be possible in your implementation to give an optional `order` parameter, e.g. `[2, 0, 1]` to align the 1st principal axis to *z*, the 2nd to *x* and the 3rd to *y*?

src/biotite/structure/transform.py
```diff
@@ -234,6 +234,78 @@ def rotate_about_axis(atoms, axis, angle, support=None):
...
+    """
+    Translate and rotate the atoms to be centered at the origin with the
+    principal axes aligned to x, y, and z.
+
```
padix-key, line 241: I think a short note about the meaning of the result could be helpful for users that do not know about principal component analysis.

```suggestion
    
    The resulting coordinates have the highest variance in the x-axis and the lowest variance on the z-axis.
```

src/biotite/structure/transform.py
```diff
@@ -10,7 +10,7 @@
 __name__ = "biotite.structure"
 __author__ = "Patrick Kunzmann"
```
padix-key, line 11: Feel free to add your name here.

src/biotite/structure/transform.py
```diff
@@ -234,6 +234,78 @@ def rotate_about_axis(atoms, axis, angle, support=None):
...
+
+    Examples
+    --------
+    Align principal components to xyz axes
+
+    >>> print("original variance =", molecule.coord.var(axis=0))
+    original variance = [ 4.8955374  2.3114026 16.263195 ]
+    >>> moved = orient_principal_components(molecule)
+    >>> print("moved variance =", moved.coord.var(axis=0))
+    moved variance = [21.425678   1.1962254  0.8482297]
```
padix-key, line 267: This doctest fails in the CI, since `molecule` is not known. However, `atom_array` is, based on PDB 1L2Y. I am not sure though, whether the resulting numbers are correct in this case.

```suggestion
    >>> print("original variance =", atom_array.coord.var(axis=0))
    original variance = [ 4.8955374  2.3114026 16.263195 ]
    >>> moved = orient_principal_components(atom_array)
    >>> print("moved variance =", moved.coord.var(axis=0))
    moved variance = [21.425678   1.1962254  0.8482297]
```

The other failing test is not linked to your PR, but is caused by a change in the PDB.

padix-key (changes requested):
Thank you for the PR. I have only a few minor suggestions and a question, that I put into the review.

claudejrogers:
Yes, I can make these changes. Meanwhile, I'll reply to some of the comments.

src/biotite/structure/transform.py
```diff
@@ -234,6 +234,78 @@ def rotate_about_axis(atoms, axis, angle, support=None):
...
+            f" and 3 dimensions, {col} given."
+        )
+
+    # place centroid of the atoms at the origin
+    M = X - X.mean(axis=0)
+
+    # iterate a few times to ensure the ideal rotation has been applied
+    identity = np.eye(3)
+    MAX_ITER = 50
+    for _ in range(MAX_ITER):
```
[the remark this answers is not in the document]
claudejrogers: I thought it wasn't necessary to loop either. While testing, I thought that (a) if the optimal rotation was found, the calculated rotation matrix should be close to `np.eye(3)` and the variance should descend from x to z. While testing, I found I needed to loop a few times for it to converge (I used the structure of cholesterol from 7VGY.pdb to test). I'll post an example.

src/biotite/structure/transform.py
claudejrogers: It should be. I'll test to confirm. Worst case, I could call `rotate_about_axis`.

src/biotite/structure/transform.py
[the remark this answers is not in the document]
claudejrogers: An example:
```python
import numpy as np
from biotite.structure.io import pdb

pdb_file = pdb.PDBFile.read("../pdbreader/7VGY.pdb")
mol = pdb.get_structure(pdb_file, model=1, include_bonds=True)
clr = mol[mol.res_name == "CLR"]

coords = clr.coord

centered = coords - coords.mean(axis=0)

identity = np.eye(3)
MAX_ITER = 50
for i in range(MAX_ITER):
    # PCA, W is the component matrix, s ~ explained variance
    _, sigma, components = np.linalg.svd(centered)

    # sort 1st, 2nd, 3rd pca compontents along x, y, z
    idx = sigma.argsort()[::-1]
    A = np.eye(3)[:, idx]

    # Run the Kabsch algorithm to orient principal components
    # along the x, y, z axes
    V, _, Wt = np.linalg.svd(components.T @ A)

    if np.linalg.det(V) * np.linalg.det(Wt) < -0:
        V[:, -1] *= -1
    # Calculate rotation matrix
    rotation = V @ Wt
    if np.isclose(rotation, identity, atol=1e-5).all():
        break
    # Apply rotation, keep molecule centered on the origin
    centered = centered @ rotation
    print(f"iter = {i + 1}\nvariance = {centered.var(axis=0)}\nrotation matrix = {rotation}")
```

Results in this output:
```python
iter = 1
variance = [ 5.21427846  1.30711532 16.94874175]
rotation matrix = [[-0.99420943 -0.10450271  0.02503573]
 [-0.0957528   0.96726584  0.23500681]
 [-0.04877506  0.23124875 -0.97167125]]
iter = 2
variance = [21.42568012  1.1962256   0.8482298 ]
rotation matrix = [[ 0.46052478  0.07427249  0.88453407]
 [ 0.07427249 -0.996223    0.04498147]
 [ 0.88453407  0.04498147 -0.46430178]]
iter = 3
variance = [21.42568012  1.1962256   0.8482298 ]
rotation matrix = [[-1.00000000e+00  0.00000000e+00  0.00000000e+00]
 [ 0.00000000e+00 -1.00000000e+00  1.11022302e-16]
 [ 0.00000000e+00  1.11022302e-16  1.00000000e+00]]
iter = 4
variance = [21.42568012  1.1962256   0.8482298 ]
rotation matrix = [[ 1.0000000e+00  0.0000000e+00  0.0000000e+00]
 [ 0.0000000e+00 -1.0000000e+00  4.4408921e-16]
 [ 0.0000000e+00 -4.4408921e-16 -1.0000000e+00]]
```

So, one rotation was not sufficient, but it converges rapidly. An issue could be symmetric molecules, like benzene, where the variance would be identical in two dimensions and zero in the third. That's why I used a for loop with a MAX_ITER.

src/biotite/structure/transform.py
claudejrogers: Ok, with an order kwarg added, this:
```python
import numpy as np
from biotite.structure.io import pdb
from biotite.structure import orient_principal_components
from biotite.structure import graphics


pdb_file = pdb.PDBFile.read("../pdbreader/7VGY.pdb")
mol = pdb.get_structure(pdb_file, model=1, include_bonds=True)
clr = mol[mol.res_name == "CLR"]

orientations = (
    np.array([0, 1, 2]),
    np.array([0, 2, 1]),
    np.array([1, 0, 2]),
    np.array([2, 0, 1]),
    np.array([2, 1, 0]),
    np.array([1, 2, 0]),
)

carbon_colors = (
    (0.2, 0.2, 0.2),
    (0.5, 0.5, 0.5),
    (0.0, 0.8, 0.0),
    (0.0, 0.4, 0.0),
    (0.0, 0.0, 0.8),
    (0.0, 0.0, 0.4),
)

fig = plt.figure(figsize=(4.0, 4.0))
ax = fig.add_subplot(111, projection="3d")

ax.plot3D([-10, 10], [0, 0], [0, 0], 'r')
ax.plot3D([0, 0], [-10, 10], [0, 0], 'g')
ax.plot3D([0, 0], [0, 0], [-10, 10], 'b')

for order, color in zip(orientations, carbon_colors):
    print(f"\norientation = {order}")
    moved = orient_principal_components(clr, order=order)
    colors = np.zeros((clr.array_length(), 3))
    colors[clr.element == "O"] = (0.8, 0.0, 0.0)
    colors[clr.element == "C"] = color
    print(moved.coord.var(axis=0).argsort(), order)
    graphics.plot_atoms(ax, moved, colors, line_width=3, background_color="white", zoom=1.5)
plt.show()
```

Produces this:
CLR_figure

src/biotite/structure/transform.py
```diff
@@ -234,6 +234,105 @@ def rotate_about_axis(atoms, axis, angle, support=None):
...
+
+    Parameters
+    ----------
+    atoms : AtomArray or ndarray, shape=(n,3)
+        The atoms of which the coordinates are transformed.
+        The coordinates can be directly provided as :class:`ndarray`.
+    order : array-like, length=3
+        The order of decreasing variance. Setting order to (2, 0, 1)
+        results in highest variance along the y-axis and lowest along
+        the x-axis. Default = (0, 1, 2)
```
padix-key, line 257: I added some highlighting to the `order` parameter and changed literal code to monospace. Furthermore, I think *'parameter'* is the more fitting word here, since *'keyword'* indicates that the parameter can exclusively given as keyword and not as positional parameter (as in `**kwargs`), in my opinion.

```suggestion
    Translate and rotate the atoms to be centered at the origin with
    the principal axes aligned to the Cartesian axes, as specified by
    the `order` parameter. By default, x, y, z.

    By default, the resulting coordinates have the highest variance in
    the x-axis and the lowest variance on the z-axis. Setting the `order`
    parameter will change the direction of the highest variance.
    For example, ``order=(2, 1, 0)`` results in highest variance along the
    z-axis and lowest along the x-axis.

    Parameters
    ----------
    atoms : AtomArray or ndarray, shape=(n,3)
        The atoms of which the coordinates are transformed.
        The coordinates can be directly provided as :class:`ndarray`.
    order : array-like, length=3
        The order of decreasing variance. Setting `order` to ``(2, 0, 1)``
        results in highest variance along the y-axis and lowest along
        the x-axis. Default = ``(0, 1, 2)``.
```

src/biotite/structure/transform.py
```diff
@@ -234,6 +234,105 @@ def rotate_about_axis(atoms, axis, angle, support=None):
...
+    >>> print("original variance =", atom_array.coord.var(axis=0))
+    original variance = [26.517471 20.008837  9.324988]
+    >>> moved = orient_principal_components(atom_array)
+    >>> print("moved variance =", moved.coord.var(axis=0))
+    moved variance = [28.906227 18.495403  8.449674]
+    >>> # Note the increase in variance along the x-axis
+    >>> # Specifying the order keyword changes the orientation
+    >>> moved_z = orient_principal_components(atom_array, order=(2, 1, 0))
+    >>> print("moved (zyx) variance =", moved_z.coord.var(axis=0))
+    moved (zyx) variance = [ 8.449674 18.495403 28.906227]
```
padix-key, line 284: For the doctests, the values are rounded to 3 decimal places. With the rounded values your output passes the test.

```suggestion
    >>> print("original variance =", atom_array.coord.var(axis=0))
    original variance = [26.517 20.009  9.325]
    >>> moved = orient_principal_components(atom_array)
    >>> print("moved variance =", moved.coord.var(axis=0))
    moved variance = [28.906 18.495  8.450]
    >>> # Note the increase in variance along the x-axis
    >>> # Specifying the order keyword changes the orientation
    >>> moved_z = orient_principal_components(atom_array, order=(2, 1, 0))
    >>> print("moved (zyx) variance =", moved_z.coord.var(axis=0))
    moved (zyx) variance = [ 8.450 18.495 28.906]
```

padix-key (changes requested):
Thank you for the changes. I just suggested some changes to the docstring and fixed the doctest.
If you like, you can also add yourself to the `CONTRIB.rst`.

padix-key (approved):
The new functionality now looks fine to me. Thank you for the contribution!

PR #377, opened 2022-03-22
Ensure align_vectors inputs are correct

claudejrogers:
If origin_direction or target_direction are shape (1, 3), it is
possible to raise a DeprecationWaring when declaring v_c.
I addressed this by adding a check of the shape of these inputs.
End users could call np.squeeze to ensure the vectors have the
correct shape.

If origin_direction or target_direction have some other shape, the
rest of the code won't work anyway.


padix-key (changes requested):
Thank you for the fix. You are right, retrospectively I see no reason why I initially used an ellipsis in the matrix. I just fixed some typos in the PR. After your approval the PR is ready to be merged from my side.

padix-key (approved):

PR #382, opened 2022-04-03
Cleaned up superimpose code

claudejrogers:
Removed duplicate code handling different cases with AtomArray versus AtomArrayStack input.
Fixed random bug in geometry.py that would have affected error handling.

### DIFF
--- a/src/biotite/structure/geometry.py
+++ b/src/biotite/structure/geometry.py
@@ -632,7 +632,7 @@ def _call_non_index_function(function, expected_amount,
     if indices.shape[-1] != expected_amount:
         raise ValueError(
             f"Expected length {expected_amount} in the last dimension "
-            f"of the indices, but got length {pairs.shape[-1]}"
+            f"of the indices, but got length {indices.shape[-1]}"
         )
     coord_list = []
     for i in range(expected_amount):
--- a/src/biotite/structure/superimpose.py
+++ b/src/biotite/structure/superimpose.py
@@ -152,7 +152,7 @@ def superimpose(fixed, mobile, atom_mask=None):
         s_coord[i] = np.dot(rotation, s_coord[i].T).T
         transformations[i] = (-mob_centroid[i], rotation, fix_centroid)
     s_coord += fix_centroid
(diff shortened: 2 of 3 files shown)


padix-key (approved):
Thank you, the changes look good to me. Now that the function is based on coordinates, we could also think about allowing also directly coordinate `ndarray`s as parameter for `superimpose()` and `superimpose_apply()` as alternative to `AtomArray`/`AtomArrayStack`. Do you have an opinion on this and would you like to implement it in this PR or is this feature something for the future?

claudejrogers:
Yes, that is a good idea. I can do that in a later PR.

padix-key (changes requested):
Hi, I found an edge case where the code change does not work: If an input structure with only a single model or atom is given, these dimensions are also 'squeezed'. I added a test case and a fix in this review.

claudejrogers:
Good catch on the single-model AtomArrayStack.

I would think the minimum number of atoms in the inputs, or after applying a mask, should be 3. 
With 2 input atoms, the covariance matrix would have shape 2 x 2.

padix-key:
The covariance matrix should include 3x3 entries independent of the number of atoms to my understanding, since it gives the covariance for the coordinate components, i.e.
```
Cov(X,X), Cov(X,Y), Cov(X,Z)
Cov(Y,X), Cov(Y,Y), Cov(Y,Z)
Cov(Z,X), Cov(Z,Y), Cov(Z,Z)
```

padix-key:
I merge the PR then, thank you for the effort!

claudejrogers:
I'm referring to this line in `_superimpose`

```python
    # Calculate covariance matrix
    cov = np.dot(x.T, y)
```

If x and y are 2 x 3, `cov` will be 2 x 2.

Plus, "superimposing" to one or two atoms is underspecified, since there are infinite equally valid solutions.

padix-key:
I tested this line of code and it always returns a `3x3` matrix for me independent of the number of atoms, as `x` is transposed.
Apart from that I agree with you that superimposing one or two atoms onto another set of atoms does not give an unambiguous solution. Maybe a warning could be given in this scenario.

PR #401, opened 2022-06-14
fix(pdb): support pdb without charge column

f-allain:
Hi,

I have a minor change proposition. It concerns the support of empty charge column with `ATOM` records in PDB. 

## Description
We actually faced parsing issues by reading files from [`PDB-REDO` database](https://pdb-redo.eu/) where `charge` columns are empty within re-refined (only) structures like [1cbs](https://pdb-redo.eu/db/1cbs/1cbs_besttls.pdb.gz) which is still compatible with `PDB format` since this column [is marked as "optional" in the format specification](https://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ATOM). The example bellow should be enough to reproduce this issue

```shell
biotite on  remotes/origin/HEAD [$?] via biotite via 🐍 biotite 
➜ wget https://pdb-redo.eu/db/1cbs/1cbs_besttls.pdb.gz && gunzip 1cbs_besttls.pdb.gz
```

```python
>>> from biotite.structure.io.pdb.file import PDBFile
>>> pdb_file = PDBFile.read("1cbs_besttls.pdb")
>>> pdb_file.get_structure(extra_fields=["charge"])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/fallain/WIP/biotite/src/biotite/structure/io/pdb/file.py", line 391, in get_structure
    array.set_annotation("charge", np.where(
ValueError: invalid literal for int() with base 10: ''
```

## Proposal
*  Change `set_annotation` call for `charge` field (L406) in `PDBFile.get_structure` in order to init to default value for empty fields (`""` or `"  "`)

### DIFF
--- a/src/biotite/structure/io/pdb/file.py
+++ b/src/biotite/structure/io/pdb/file.py
@@ -6,22 +6,21 @@ __name__ = "biotite.structure.io.pdb"
 __author__ = "Patrick Kunzmann, Daniel Bauer, Claude J. Rogers"
 __all__ = ["PDBFile"]
 
-import warnings
+from warnings import warn
+
 import numpy as np
+
+from ....file import InvalidFileError, TextFile
 from ...atoms import AtomArray, AtomArrayStack
 from ...bonds import BondList, connect_via_residue_names
-from ...box import vectors_from_unitcell, unitcell_from_vectors
-from ....file import TextFile, InvalidFileError
-from ..general import _guess_element as guess_element
-from ...error import BadStructureError
+from ...box import unitcell_from_vectors, vectors_from_unitcell
 from ...filter import (
     filter_first_altloc,
     filter_highest_occupancy_altloc,
     filter_solvent,
 )
-from .hybrid36 import encode_hybrid36, decode_hybrid36, max_hybrid36_number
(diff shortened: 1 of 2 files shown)


padix-key:
Thanks for reporting this. Although the charge column is optional the overall line length is not (https://www.wwpdb.org/documentation/file-format-content/format33/sect1.html):

> Each line in the PDB entry file consists of 80 columns.

`1cbs_besttls.pdb` only contains 78 characters, leading to the error. Nevertheless, I think *Biotite* should also support files, that are slightly malformatted, such as PDB files from *Gromacs* with missing element column.
However, the solution proposed here is specific to the charge column and works only if the line length is 78 or shorter. For 79 the charge string would be `" "`, which is not covered in this change.
Therefore, I would propose to fix too short PDB lines in general instead by padding lines with whitespace, if they are too short. I will prepare a PR for this so you can evaluate whether this alternative is a sufficient solution for you.

f-allain:
Indeed, It's more convenient to prevent it in the way you described in the PR #402. I don't have anything else to add so I guess you can close this PR :)

Issue #409, opened 2022-07-19
Changed UniProt REST API


padix-key:
The *UniProt KB* has revamped its REST API, with a new URL and changed parameter names (https://www.uniprot.org/help/api). While fetching entries still works (for now), the doctests and tests involving `search()` fail, probably due to the new `size` option, which seems to replace the `number` option.

PR #412, opened 2022-08-03
Issue 409

MaxGreil:
Fixes issue #409 

- Changed url from `uniprot` to `uniprotkb`
- In `search()`, changed parameter `limit` to new `size`

### DIFF
--- a/src/biotite/database/uniprot/query.py
+++ b/src/biotite/database/uniprot/query.py
@@ -11,7 +11,7 @@ import abc
 from .check import assert_valid_response
 
 
-_base_url = "https://www.uniprot.org/uniprotkb/"
+_base_url = "https://rest.uniprot.org/uniprotkb/search/"
 
 
 class Query(metaclass=abc.ABCMeta):
@@ -152,7 +152,7 @@ class SimpleQuery(Query):
         return f"{self._field}:{self._term}"
 
 
-def search(query, number=10):
+def search(query, number=500):
     """
     Get all UniProt IDs that meet the given query requirements,
     via the UniProt search service.
@@ -165,7 +165,7 @@ def search(query, number=10):
         The search query.
     number : int
         The maximum number of IDs that are obtained.
-        (Default: 10)
+        (Default: 500)
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


src/biotite/database/uniprot/query.py
```diff
@@ -11,7 +11,7 @@
 from .check import assert_valid_response
 
 
-_base_url = "https://www.uniprot.org/uniprot/"
+_base_url = "https://www.uniprot.org/uniprotkb/"
```
padix-key, line 14: The new search URL is slightly different.

```suggestion
_base_url = "https://rest.uniprot.org/uniprotkb/search"
```

src/biotite/database/uniprot/query.py
```diff
@@ -189,7 +190,7 @@ def search(query, number=10):
     params = {
         'query': str(query),
         'format': 'list',
-        'limit': str(number)
+        'size': str(number)
```
padix-key, line 193: A suggestion for this PR or a future change: The new REST API allows pagination similar to the RCSB PDB. Hence, we could substitute the `number` parameter with `range` analogous to [database.rcsb.search](https://www.biotite-python.org/apidoc/biotite.database.rcsb.search.html). In the default case this would return all entries. Unfortunately, this could only be done in multiple requests as shown in https://www.uniprot.org/help/api_queries.

src/biotite/database/uniprot/query.py
```diff
@@ -165,6 +165,7 @@ def search(query, number=10):
         The search query.
     number : int
         The maximum number of IDs that are obtained.
+        (Default: 10)
```
padix-key, line 168: I would suggest to set this parameter to 500 as this seems to be the new maximum value of a single request (taken from https://www.uniprot.org/help/api_queries):

> Always use size=500 as this will provide fast performance.

padix-key (changes requested):
Thank you for the fix. The tests in `tests/database/test_uniprot.py` do not pass at this point, since the result for the `accession` field has been changed in the new API. Unfortunately I had to remove the thorough tests, including doctests and UniProt tests, from the CI due to some persistent issues with GitHub Actions. Therefore, I would ask you to run `tests/database/test_uniprot.py` and `tests/test_doctest.py` locally.

src/biotite/database/uniprot/query.py
MaxGreil: I will keep this in mind for a future change.

MaxGreil:
Hi @padix-key ,

I implemented the requested changes and also run the tests locally.

padix-key (approved):
Thank you for the changes. I would merge this PR, if you are ready.

padix-key:
I found another change in the new API: Some field names have changed or have been added (https://www.uniprot.org/help/query-fields). This includes for example `taxonomy_name`. If such field is selected for a `SimpleQuery`, a exception is raised, since the field name is checked against the `_fields` list.

MaxGreil:
@padix-key , I applied the changes to the query fields.

From my side this PR can now be merged.

PR #418, opened 2022-08-31
Mol2, SDF and XYZ File Parsers

entropybit:
The file parsers for those 3 files now essentially work. 
There is some fine work missing, especially with the hybridization heuristics in mol2 and auto detection
of atom_names vs elements in mol2 (one of the reasons why right now mol2 doesn't work with the amino acid test cases).
But reading small molecules, also with multiple models, should work and I have added some more files in all 3 formats for testing this.


padix-key:
Thanks for the PR. Prior to review, could you resolve the conflicts with the upstream branch?

entropybit:
Only took like a week more then expected -.- ....

src/biotite/structure/io/mol2/file.py
```diff
@@ -0,0 +1,990 @@
...
+            else:
+                if len(atom_bonds) == 3:
+                    return "C.1"
+                elif len(atom_bonds) == 2:
+                    return "C.2"
+                elif len(atom_bonds) == 1:
+                    return "C.3"                         
+    #            else:
+    #                msg = "No supported sybyl Atom type for Atom " + str(atom)
+    #                raise ValueError(msg)
```
padix-key, line 170:
```suggestion             
```

src/biotite/structure/io/mol2/file.py
```diff
@@ -0,0 +1,990 @@
...
+    elements_twoLetters = [
+        "LI", "NA", "MG", "AL",
+        "SI", "CA", "CR", "MN",
+        "FE", "CO", "CU", "CL",
+        "BR", "ZN", "SE", "MO",
+        "SN", "AR"
+    ]    
+    
+    @staticmethod        
+    def get_sybyl_atom_type(atom, bonds, atom_id):
```
padix-key, line 126: Is this function meant to be private?

For performance reasons, I would suggest to give the function all bonds, element, and the atom index directly. The bonds for each atom can be more efficiently obtained via `get_all_bonds()`.

src/biotite/structure/io/mol2/file.py
```diff
@@ -0,0 +1,990 @@
...
+                heuristic.
+                
+            atom_id: int
+                The id of the current atom that is used in the bond list.                    
+                
+        Returns
+        -------     
+            sybyl_atom_type : str
+                The name of the atom based on hybridization and element according
+                to the sybyl atom types.
```
padix-key, line 153: The indentation is wrong.

src/biotite/structure/io/mol2/file.py
```diff
@@ -0,0 +1,990 @@
...
+        "SI", "CA", "CR", "MN",
+        "FE", "CO", "CU", "CL",
+        "BR", "ZN", "SE", "MO",
+        "SN", "AR"
+    ]    
+    
+    @staticmethod        
+    def get_sybyl_atom_type(atom, bonds, atom_id):
+        """
+        This (horible horribel) function is meant to translate all occuring
```
padix-key, line 128: I know what you mean :smile:, but for the purpose of a docstring, I think it is better to remove the textual representation of the emotions.

src/biotite/structure/io/mol2/file.py
```diff
@@ -0,0 +1,990 @@
...
+        if atom.element == "C":                          
+            if 5 in types:
+                return "C.ar"
+            else:
+                if len(atom_bonds) == 3:
+                    return "C.1"
+                elif len(atom_bonds) == 2:
+                    return "C.2"
+                elif len(atom_bonds) == 1:
+                    return "C.3"
```
padix-key, line 167: I think the hybridization is wrong here: For example, if the carbon has two bonds it should be sp (C.1). The case of one bond should in theory never occur.

padix-key (changes requested):
Hi, here is the first part of my review. Some general remarks:
- I think for consistency reasons the `get_structure()` methods should take the `model` parameter analogous to for example `PDBFile` and select the return type based on it.
- There are some commented out lines of code that remain.
- Sometimes, there are (multiple) empty lines, without obvious structuring reason.
- You can add yourself to `CONTRIB.rst` via this PR, if you like.

src/biotite/structure/io/mol2/file.py
entropybit: As discussed elsewher this should remain a static function. I renamed it to
**transcribe_sybyl_atom_type** to make clear that this is not a getter function

src/biotite/structure/io/mol2/file.py
entropybit: Should be corrected now.

src/biotite/structure/io/mol2/file.py
entropybit: Jup, sorry. You are right of course, comment edited accordingly.

src/biotite/structure/io/mol2/file.py
entropybit: I corrected this but I have to say I really do not understand this. Based on the different hybridization being determined by number of binding, I should get N_bonds==2, N_bonds==3, N_bonds==4 in the part of the heuristic where "N" is being processed. However if I do that  some of the test files already have wrong atom_type entries.

PR #422, opened 2022-09-07
fix(pdbparsing): parse "+X" and "X+" charges

### LINKED ISSUE #421: PDB parsing fails if the charge is specified as "+X" instead of "X+"
Hi!

Thanks for your great work! Here's a small bug report:

**Context**

I have a problem parsing PDB files that were apparently written with [REFMAC 5.8.0258](https://www.ccp4.ac.uk/html/refmac5.html).

**Symptom**

After [this file](https://gist.github.com/aurelg/27a08f838d81a3e11893c503def84e16) is loaded (as `pdb_file`), a call to `.get_structure(altloc="all", extra_fields=["charge"])` fails.

Ie. the following code:

```python
pdb.PDBFile.read("1atp_charged.pdb").get_structure(
    altloc="all",
    extra_fields=["charge"],
)
```

leads to the following error:

```
Traceback (most recent call last):
  File "/YYYYYYY", line 5, in <module>
    pdb.PDBFile.read("1atp_charged.pdb").get_structure(
  File "/XXXXXX/biotite/structure/io/pdb/file.py", line 400, in get_structure
    "charge", np.where(charge == "  ", "0", charge).astype(int)
ValueError: invalid literal for int() with base 10: '2+'
```

**Root cause analysis**

The [PDB format specification](https://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ATOM) explicitly state that `Columns 79 - 80 indicate any charge on the atom, e.g., 2+, 1-. In most cases, these are blank.`.

Biotite consequently reverses the charge column before parsing in [io/pdb/file.py:370](https://github.com/biotite-dev/biotite/blob/master/src/biotite/structure/io/pdb/file.py#L370):

```python
charge_raw[i] = line[_charge][::-1]
```

Unfortunately, REFMAC (and possibly others?) seems to store charges as follows (notice the `+2` instead of `2+`, [here](https://gist.github.com/aurelg/27a08f838d81a3e11893c503def84e16#file-1atp_charged-pdb-L3494-L3495)):

```
HETATM 2937 MN    MN E 351      14.676   8.154  -3.730  1.00 23.67          MN+2
HETATM 2938 MN    MN E 352      13.555   6.903  -0.131  1.00 21.04          MN+2
```

This notation violates the PDB format specification and breaks parsing with biotite as the reversed charge columns `2+` cannot be converted to an integer, hence the stack trace above. 

**Possible mitigation and corresponding motivation**

Adding support for non-standard PDB files might be beyond the scope of biotite. However, in this particular case, the fix should have no side effect and would allow biotite users to load their files.

The line `io/pdb/file.py:370` 

```python
charge_raw[i] = line[_charge][::-1]
```

Would have to be replaced by:

```python
if line[_charge][0] in "+-":
  charge_raw[i] = line[_charge]
else:
  charge_raw[i] = line[_charge][::-1] # turn "1-" into "-1"
```

If you agree, I can make a PR?

aurelg:
Fixes #421

### DIFF
--- a/src/biotite/structure/io/pdb/file.py
+++ b/src/biotite/structure/io/pdb/file.py
@@ -367,7 +367,11 @@ class PDBFile(TextFile):
             element[i] = line[_element].strip()
             altloc_id[i] = line[_alt_loc]
             atom_id_raw[i] = line[_atom_id]
-            charge_raw[i] = line[_charge][::-1]  # turn "1-" into "-1"
+            # turn "1-" into "-1", if necessary
+            if line[_charge][0] in "+-":
+                charge_raw[i] = line[_charge]
+            else:
+                charge_raw[i] = line[_charge][::-1] 
             occupancy[i] = float(line[_occupancy].strip())
             b_factor[i] = float(line[_temp_f].strip())
         


padix-key:
Hi, this PR fixes a rather special case, so I ran some simple benchmarks to see if this change has a performance impact.
Since I found no notable changes in run time, this change is fine for me. Thank you for providing the fix!

PR #423, opened 2022-09-16
support for Path objects

edikedik:
`Path` objects provide an OS-agnostic interface to the filesystem and are compatible with the built-in `open` function. However, using them in biotite raises an error and requires an explicit opening externally. I believe they should be handled the same way as the `str` in this case.

### DIFF
--- a/src/biotite/file.py
+++ b/src/biotite/file.py
@@ -9,6 +9,8 @@ __all__ = ["File", "TextFile", "InvalidFileError"]
 import abc
 import io
 import warnings
+from pathlib import Path
+
 from .copyable import Copyable
 import copy
 
@@ -50,7 +52,6 @@ class File(Copyable, metaclass=abc.ABCMeta):
             representing the parsed file.
         """
         pass
-        
             
     def _deprecated_read(self, file, *args, **kwargs):
         """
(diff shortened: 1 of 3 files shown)


padix-key:
Hi, thanks for the PR. I just had a look at the `open()` function, which in general accepts a path-like object:

> An object representing a file system path. A path-like object is either a [str](https://docs.python.org/3/library/stdtypes.html#str) or [bytes](https://docs.python.org/3/library/stdtypes.html#bytes) object representing a path, or an object implementing the [os.PathLike](https://docs.python.org/3/library/os.html#os.PathLike) protocol.

Since `Path` implements `PathLike`, I think it would be more general to write
```
if isinstance(file, str) or isinstance(file, bytes) or isinstance(file, PathLike):
```
or in shorter form
```
if isinstance(file, (str, bytes, PathLike)):
```

For more consistency, it would be nice to have this `Path` support also for the binary `biotite.structure.io.mmtf.MMTFFile`.
Could you add these changes to the PR? If you like you can also add yourself to `CONTRIB.rst`.

padix-key:
Is this PR ready to be merged?

edikedik:
> Is this PR ready to be merged?

Yes.

Issue #424, opened 2022-09-16
fixed list of amino acid three-letter codes may lead to unexpected results


edikedik:
Hi,

The function `filter_amino_acids` utilizes a hardcoded list of residues corresponding to the canonical alphabet. 

```python
_ext_aa_list = ["ALA","ARG","ASN","ASP","CYS","GLN","GLU","GLY","HIS","ILE",
                "LEU","LYS","MET","PHE","PRO","SER","THR","TRP","TYR","VAL",
                "MSE", "ASX", "GLX", "SEC", "UNK"]
```

Although it seems sane, it may lead to unexpected results when non-canonical residues reside within the protein sequence.
Indeed, in this case, function wrongly makes the protein chain discontinuous likely causing downstream errors.

I see three possible solutions.

1. Rename the function! For instance, `filter_canonical_amino_acids` clearly states the functionality.
2. Just extend the list and allow the user to input a custom one. For instance, residues such as  "TPO", "PTR", and "SEP" are phosphorylated threonine, tyrosine, and serine. These are quite common in PDB. There are many other possible covalent modifications, either crafted by authors to improve the molecule's stability or added naturally as part of post-translational modifications. Of course, this makes it notoriously hard to account for all of them, yet it doesn't mean nothing should be done.
3. Extend/improve the function or create a new one. Although the current documentation is correct, the function may be wrongly perceived as the filter of amino acid residues **that belong to a protein chain**. Indeed, I suspect you planned it as such, as you rely on it in the `filter_backbone`. However, there are cases when amino acids are used as ligands. Thus, an extension/improvement, I believe, should rely on both protein continuity and maybe utilize the extended list I suggested above. For instance, an empirical workflow is to break the chain into continuous pieces and treat any residue bounded by  residues from the hardcoded list (such as `_ext_aa_list`) as an amino acid.

Let me know what you think. If you agree that, at least, there is a problem, and we come to some consensus, I could implement it via PR. Thanks.

tomtomhdx:
In general I agree with @edikedik.

I would propose to use the `_chem_comp.type` property from the  _PDB Chemical Component Dictionary_ as it is already done in `filter_nucleotides()`. I would be happy to implement that as I already wrote the relevant scripts for `filter_nucleotides()`.

Furthermore, as @edikedik suggested we could add a `filter_canonical_amino_acids()` and for completeness `filter_canonical_nucleotides()`.

padix-key:
[...]

PR #425, opened 2022-09-17
Improve filters for nucleotides, amino acids and add filter for carbohydrates.

tomtomhdx:
This PR implements part of the changes discussed in #424:

- `filter_amino_acids()` now uses an extended amino acid list extracted from the _PDB Chemical Component Dictionary_, where the `_chem_comp.type` property corresponds to peptide monomers.
- The nucleotide list for `filter_nucleotides()` was updated using the current version of the _PDB Chemical Component Dictionary_.
- `filter_carbohydrates()` filters for residues, where the `_chem_comp.type` corresponds to polysaccharide monomers.
- `filter_canonical_amino_acids()` filters for the 22 proteinogenic amino acids.
- `filter_canonical_nucleotides()` filters for the canonical DNA/RNA nucleotides A, C, G, T/U.
- Analogous to `nucleotide_names()`, `carbohydrate_names()` and `amino_acid_names()` return a list of all monomers that correspond to polysachharide and peptide monomers respectively.

I am unsure why `filter_amino_acids()` checks that the `res_id` is not -1:
`return ( np.isin(array.res_name, _amino_acid_list) & (array.res_id != -1) )`


padix-key (approved):
Thank you for the implementation, the changes look good to me.

As final remark I would suggest writing a second sentence in the `filter_xxx()` docstrings that mentions the origin from the CCD and the included `_chem_comp.type`.

tomtomhdx:
> Thank you for the implementation, the changes look good to me.
> 
> As final remark I would suggest writing a second sentence in the `filter_xxx()` docstrings that mentions the origin from the CCD and the included `_chem_comp.type`.

Good Idea! I added the `_chem_comp.type` values to the respective _Notes_ section of the functions.

padix-key:
Thanks. Can i merge this PR?

tomtomhdx:
Yes its ready from my side.

padix-key:
> I am unsure why filter_amino_acids() checks that the res_id is not -1:
return ( np.isin(array.res_name, _amino_acid_list) & (array.res_id != -1) )

It seems like I overlooked this part of your comment. I think I added this condition, since `res_id` is often `-1` in mmCIF files, if a residue is not part of a chain. As this is not true for all file formats and sometimes amino acids inside a chain also get such residue ID, I would propose removing this extra condition.

tomtomhdx:
I completely forgot about that! I will remove the extra condition then.

PR #430, opened 2022-09-23
Refactoring `check_bond_continuity`

edikedik:
Hi,

The current `check_bond_continuity()` is specialized towards backbone atoms, which is unintuitive. I propose a broader version of this function doing the same job without implicit filtering for the backbone atoms.

This change also simplifies the calculation (no need for masking). Furthermore, what the function was doing is now encapsulated into `check_backbone_bond_continuity()` which naturally relies on `check_bond_continuity()`. 

EDIT: The above claim was refuted by reality: I found no better way to calculate this based on indices. Hence, the PR is basically some refactoring. Considering this, it only makes sense if you agree with the first proposition regarding the function clarification. In that case, @padix-key please let me know if I should finalize the code with comments/etc.


padix-key:
The name `check_bond_continuity()` originates from a time, when Biotite was primarily used for protein structure analysis and the `BondList` did no exist. So I am in favor of refactoring and generalizing this functionality in the way you did. However, if we are renaming the function, we could discuss giving it an even more clear name:

- `backbone` is somewhat ambiguous, since that name could also refer to nucleic acids (in consequence this would also be true for `filter_backbone()`, but I would keep that function name at least for now)
- `bond` indicates that a `BondList` is used, which is not the case

At this point I have no complete idea for a new clear and short name, yet. So I am open for suggestions. Maybe `backbone` could be replaced by `peptide`.

edikedik:
The `check_bond_continuity()` function essentially checks for continuity in a consecutive list of atoms, as opposed to the `BondList` that recognizes the non-linear nature of a molecule. Therefore, I would propose renaming it into `check_linear_continuity()` or `check_linear_integrity()`. 

As for the "backbone" function, it already implies a linear, non-branching structure in the case of both nucleic acids and peptides. Thus, I would suggest using `check_backbone_continuity()` with a keyword argument `backbone_type` evaluating to either "peptide" or "nucleotide," the former being default. That naming, in my view, combined with clear docs mentioning `BondList` somewhere, should do the job.

padix-key:
I think `check_linear_continuity()` and `check_backbone_continuity()` are good solutions then. For more convenience I propose to simply check for both, nucleotide and peptide discontinuities without additional parameter, where the transition from peptide to nucleotide and vice versa would be always discontinuous.

For nucleotides the corresponding atom names would be `P`, `O5'`, `C5'`, `C4'`, `C3'` and `O3'`. I think the order of the atoms should also be correct, so the principle of `check_linear_continuity()` should also work here.

I would also suggest keeping the old `check_bond_continuity()` for now, let it call `check_backbone_continuity()` and raise a `DeprecationWarning` as well as mention the deprecation in the docstring as done e.g. [here](https://github.com/biotite-dev/biotite/blob/master/src/biotite/temp.py#L35-L63).

edikedik:
I finally got back to it.

To summarize the changes I've implemented:
- Clarified `filter_backbone()`: now we have `filter_peptide_backbone()` and `filter_nucleotide()` backbone.
- Added `check_linear_bond_continuity()`. In contrast to the `check_bond_continuity()`, it points to atoms for which the __next__ atom is discontinuous.
- Clarified `check_backbone_continuity()` according to our discussion above. Now it relies on the filter functions above.

PR #439, opened 2022-11-15
Add rounding to newick strings

tomtomhdx:
Some programs support the display of edge labels on trees imported using the newick format. Floating point numbers can be unnecessarily precise resulting in inconvenient representations.

This PR adds the argument `round_distance` to the `to_newick()` method of `biotite.sequence.phylo::Tree`. It allows to round the edge weights in the newick representation to a given precision.


padix-key (changes requested):
I think these changes are a good improvement. Could you add a test to `tests/sequence/test_phylo.py` that checks whether rounding (and omitting rounding) works for a simple tree with known Newick string?

padix-key (approved):
Thanks for adding the test.

PR #447, opened 2022-12-08
Adding a temp_dir parameter to the VinaApp

entropybit:
Using the VinaApp for docking creates temporary files. In the original implementation those are just stored in /tmp which usually is fine. However, if one wants to use VinaApp for (large scale) docking on a cluster there might not be a /tmp folder or the path to it might be somewhere else.  Usually a distributed file system is available as scratch which is perfectly fine with handling a lot of short living files. Therefore, one might want to or even need to have control over where the tempory files are created.  I propose to simply add a parameter temp_dir that defaults to None which we will set the NamedTemporaryFiles parameter dir to. If dir is set to None it will just create a file under /tmp otherwise the given path is used. This can be tested by doing:

```python
x=NamedTemporaryFile()
y=NamedTemporaryFile(dir=None)
z=NamedTemporaryFile(dir=".")
print(x.name)
print(y.name)
print(z.name)
```
which will create 3 different termporary files and print their full path.

### DIFF
--- a/CONTRIB.rst
+++ b/CONTRIB.rst
@@ -11,3 +11,4 @@ CONTRIBUTORS
 - Claude J. Rogers <https://github.com/claudejrogers>
 - Ivan Reveguk <https://github.com/edikedik>
 - Jan H. Krumbach <https://github.com/JHKru>
+- Benjamin E. Mayer <https://github.com/entropybit>
--- a/src/biotite/application/autodock/app.py
+++ b/src/biotite/application/autodock/app.py
@@ -45,6 +45,8 @@ class VinaApp(LocalApp):
         By default, the receptor has no flexibility.
     bin_path : str, optional
         Path to the *Vina* binary.
+    temp_dir : str, optional
+        Path that will be used instead of /tmp for creating temporary files.
 
     Examples
     --------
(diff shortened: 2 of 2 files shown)


padix-key:
Hi, thanks for the suggestion. However I have a few issues with this proposed change.

- Python's temporary files should work on most systems, even if `/tmp` does not exist: (https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir). Hence, this is quite a special problem.
- This change would not only be required in `VinaApp` but all `LocalApp` classes in Biotite for consistency. We could add a common method in the `LocalApp` superclass for this purpose, but still some changes would be needed in all `LocalApp` subclasses.

Still I think that your case is valid. Python has the `tempfile.tempdir` attribute that is able to globally set the default temporary directory:
```python
>>> import tempfile
>>> tempfile.tempdir = "/some/directory"
>>> f = tempfile.NamedTemporaryFile()
>>> print(f.name)
'/some/directory/tmpbbfu6jqt'
```
Note that `"/some/directory"` must exist prior to creating the `NamedTemporaryFile`.
Does this alternative solution work for you?

entropybit:
Yeah, seems like this worked. And also it apparently was not the problem. The run I just tried gave me 
```
OSError: [Errno 24] Too many open files: '/work/scratch/b_mayer/dna_pkcs_docking/tmp/tmpwylhka5t.pdbqt'
```
I thought this might dissapear if I do it on the GPFS instead of /tmp (apparently that also existed sorry about that).
However, I did not consider the implicatiosn for the App and thought that this would only be a small change for the VinaApp.
I think I will close with this ^^.

PR #451, opened 2023-01-04
Refactored CI with Python 3.11 support

danpf:
I was having trouble installing a package and biotite was a dependency but it has numpy limited to 1.21. but 1.23 is required for python 3.11

this pr lifts that limit to 1.24.  I also made some changes to the testing matrix for github actions to reflect these changes.


padix-key:
Thanks for the adaptions. I will have a deeper look at your proposed version changes tomorrow.

.github/workflows/test.yml
```diff
@@ -7,35 +7,8 @@ jobs:
...
-            numpy: '1.16'
-          - os: windows-latest
-            python: '3.9'
-            numpy: '1.19'
-          - os: windows-latest
-            python: '3.10'
-            numpy: '1.21'
-
+        os: ['ubuntu-latest', 'windows-latest', 'macos-latest']
+        pyrestrictions: [{python: '3.8', numpy: '1.19'}, {python: '3.10', numpy: '1.21'}, {python: '3.11', numpy: '1.23'}]
```
padix-key, line 11: I like the organization using the Cartesian product of OS and version, it looks much cleaner to me. However I think the key name `versions` would be more descriptive and shorter than `pyrestrictions`. Furthermore I think you wanted to write Python 3.9 rather than 3.8.
```suggestion
        versions: [{python: '3.9', numpy: '1.19'}, {python: '3.10', numpy: '1.21'}, {python: '3.11', numpy: '1.23'}]
```

setup.py
```diff
@@ -106,7 +106,7 @@ def get_extensions():
     },
     
     install_requires = ["requests >= 2.12",
-                        "numpy >= 1.19",
+                        "numpy >= 1.19, <= 1.24",
```
padix-key, line 109: The upper version boundary is not necessary here, since more modern versions of NumPy should also work. The reason for the upper bound is `pyproject.toml`, is that a NumPy version as low as possible allows compilation of modules that work for that NumPy version and all versions above.

```suggestion
                        "numpy >= 1.19",
```

padix-key (changes requested):
The changes look largely good to me. However the version changes in `test.yml` should also be applied to `deploy.yml`. I think the Cartesian product of OS and version is not applicable there (at least not that simple), as the `source` attribute is also present there.

.github/workflows/test.yml
danpf: I just did 3.8 because that was the minimum version you had tested previously. I'll update it though.

.github/workflows/deploy.yml
```diff
@@ -8,44 +8,23 @@ jobs:
...
+          - {python: '3.9', numpy: '1.19'}
+          - {python: '3.9', numpy: '1.23'}
+          - {python: '3.9', numpy: 'none'}
+
+          - {python: '3.10', numpy: '1.21'}
+          - {python: '3.10', numpy: '1.23'}
+          - {python: '3.10', numpy: 'none'}
+
+          - {python: '3.11', numpy: '1.23'}
+          - {python: '3.11', numpy: 'none'}
```
padix-key, line 27: Thanks for the changes. In this setup `source` would always be `false`, right? If this is the case, the source distribution would not be built.

Second, I do not quite understand the purpose for multiple NumPy versions for each Python version. To my understanding the lowest NumPy version should be compatible with all minor versions above, correct? So the lowest possible version should be sufficient here.

Apart from that, you may also add yourself to `CONTRIB.rst`, if you like.

danpf: hey @padix-key sorry I just was in the middle of editing and i got pulled away and forgot to finish this...  this isn't done yet. I'll address your concerns later today.

padix-key: OK, then I will review the changes, when you are ready. You do not need to hurry.

danpf:
Hey @padix-key 

Hopefully these changes aren't too drastic for you - I think it should make maintaining this much easier.

So a few things I did here:
1. Changed your test & deploy workflow to use `cibuildwheel` and `oldest-supported-numpy`
`cibuildwheel` is pretty cool since it manages knowing all of the compiled wheel versions to build + tests for you + makes deploy-ready wheels

`oldest-supported-numpy` actually manages all of the python / numpy version matching code that you were previously implementing.  This way as more python versions are added, you won't have to manually update this package.

2.  I added a pypi CD step.  Basically if you integrate a pypi secret as described https://cibuildwheel.readthedocs.io/en/stable/deliver-to-pypi/ the only thing you will need to do to post a new version to pypi is to use the github gui to create a new release.

3. During the test process I added `durations=50` to the pytest call.  I didn't dig down into the details, but for some reason some of the tests were taking a very long time (~1 hr) to run.
  * I fixed one -- but I'm not sure why it was happening.  It looked like python was treating the hetero list equality check as some strange and complicated string comparison.  The test in question is `tests/structure/test_mmtf.py` -> `test_pdbx_consistency` -> `5eil`.
 
4. I also noticed on some platform / python combinations the test `tests/sequence/test_phylo.py` -> `test_newick_rounding` so I Xfailed it.  However, you might want to just have it run exclusively on platforms/pythons you know it will work on so when it does fail, github actions catches it and reports it.

the only thing I wanted to mention is that python 3.6 end of life was ~1 year ago, and python 3.7s EOL is in 6 months.  So I would consider dropping support for 3.6 now, and 3.7 when the time comes. 

Assuming this final set of tests succeeds I think that's all I wanted to do. let me know if you have any questions.

padix-key:
Hi, thank you for the effort, your changes look promising. Since some of the elements of the change are new to me, I will probably need until next week for the review.

I will also have a look at the test issues. I already recognized that some tests take a long time, however I was not able to reproduce this issues at my home setup. Printing the duration of the slowest tests is definitely helpful here.
The `test_newick_rounding()` test should probably adapted to be able to run on all platforms.

Regarding the version requirement I would agree to bump the minimum Python version to 3.7 (and accordingly the minimum NumPy version). Could you do that in this PR?

t0mdavid-m:
Hello,
I agree that `test_newick_rounding()` should be adapted to work on all platforms. Since I wrote the test I would be happy to look into it and adapt it accordingly.

danpf:
@t0mdavid-m I just looked at the logs a little harder and I think it was because it was building for the 32 bit architecture.  I'll try reinstating it now that I've blocked 32 bit builds.
[tmp.txt](https://github.com/biotite-dev/biotite/files/10413946/tmp.txt)

danpf:
confirmed green checkmark for tests + wheels built for 24 platforms : D
https://github.com/danpf/biotite/actions/runs/3913622212

t0mdavid-m:
Very Cool!

.github/workflows/test_and_deploy.yml
```diff
@@ -0,0 +1,137 @@
...
+  upload_to_pypi_if_release:
+    needs: [test-biotite-and-build-wheels, make_sdist, test-muscle5]
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/download-artifact@v3
+      with:
+        name: artifact
+        path: dist
+
+    - name: ls dir
```
padix-key, line 130:
```suggestion
    - name: List distributions to be uploaded
```

.github/workflows/test_and_deploy.yml
```diff
@@ -0,0 +1,137 @@
...
+
+  make_sdist:
+    name: Make SDist
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+        submodules: true
+    - name: Build SDist
```
padix-key, line 115:
```suggestion
    - name: Build source distribution
```

.github/workflows/test_and_deploy.yml
```diff
@@ -0,0 +1,137 @@
...
+        run: conda install -c bioconda "muscle=5"
+      - name: Building distribution
+        run: pip wheel --no-deps -w dist .
+      - name: Installing distribution
+        run: pip install .//dist//*.whl pytest
+      - name: Testing code
+        run: mkdir tmp && cd tmp && pytest ..//tests//application//test_msa.py --durations=10 && cd .. && rm -r tmp
+
+  make_sdist:
+    name: Make SDist
```
padix-key, line 108:
```suggestion
    name: Build source distribution
```

.github/workflows/test_and_deploy.yml
```diff
@@ -0,0 +1,137 @@
...
+      with:
+        fetch-depth: 0
+        submodules: true
+    - name: Build SDist
+      run: pipx run build --sdist
+    - uses: actions/upload-artifact@v3
+      with:
+        path: dist/*.tar.gz
+
+  upload_to_pypi_if_release:
```
padix-key, line 121: I think a verbose name would be nice here to improve the graph of the action.
```suggestion
  upload_to_pypi_if_release:
    name: Upload to PyPI
```

padix-key (changes requested):
Here are some remarks from my side. Overall I like the new build system a lot and I learned a bunch of new things. I was ill last week, so I am sorry it took longer than expected to take care of this contribution.

Here are some general questions:
- On the repository of `cibuildwheel` the authors note potential license issues due to bundled linked libraries. Did you run `auditwheel show` to display whether unexpected libraries are included?
- What is the advantage of `pipx` over `pip` in the context of this action?

.github/workflows/test_and_deploy.yml
```diff
@@ -0,0 +1,137 @@
...
+  release:
+    types:
+    - published
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+env:
+  CIBW_TEST_REQUIRES: pytest
```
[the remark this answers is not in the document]
danpf: I see, sorry I don't use conda too much for python environment management so I missed that

If this is the case then we could consider setting optional dependencies in the pyproject.toml file?
https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies

I just committed an example which should work and install them via `pip install "biotite[test]"`

.github/workflows/test_and_deploy.yml
```diff
@@ -0,0 +1,137 @@
...
+        if: runner.os == 'Linux'
+        uses: docker/setup-qemu-action@v2
+        with:
+          platforms: all
+
+      - name: Build & (optionally) Test wheels
+        # MAKE SURE THIS STAYS IN SYNC WITH THE UPPER pipx call to cibuildwheel
+        uses: pypa/cibuildwheel@v2.11.4
+        with:
+          only: ${{ matrix.only }}
```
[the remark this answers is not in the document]
danpf: yes it is slightly different.
https://github.com/pypa/cibuildwheel/blob/main/action.yml

danpf:
@padix-key 
I want to mention that the reason we need to specify 'cibuildwheel' versions in two places is that the building+tests took so long on other platforms that the time limit for free github actions was hit.   Because it took so long I had to use the 'create a matrix separately' method.  If the build & test times can be decreased then we won't need that step.


> On the repository of cibuildwheel the authors note potential license issues due to bundled linked libraries. Did you run auditwheel show to display whether unexpected libraries are included?

I have not but I can look into it

> What is the advantage of pipx over pip in the context of this action?

pipx creates isolated environments for specific apps/scripts.  It basically makes sure that when you install cibuildwheel, that it does not impact your dev environment.

danpf:
the results of `auditwheel show` are listed here:

```
biotite-0.35.1-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
is consistent with the following platform tag:
"manylinux_2_17_x86_64".

The wheel references external versioned symbols in these
system-provided shared libraries: libc.so.6 with versions
{'GLIBC_2.2.5', 'GLIBC_2.14'}

This constrains the platform tag to "manylinux_2_17_x86_64". In order
to achieve a more compatible tag, you would need to recompile a new
wheel from source on a system with earlier versions of these
libraries, such as a recent manylinux image.
```

danpf:
will see if there's a way to install mdtraj for tests - however it is worth mentioning that the mdtraj project is no longer being actively developed.

https://github.com/mdtraj/mdtraj

danpf:
Unfortunately mdtraj does not build in python 3.11 and does not appear to work on windows without explicitly installing zlib.

I think the best course of action is to lump it in with the muscle5 test and use that as the place to test the more nuanced parts of biotite

.github/workflows/test_and_deploy.yml
```diff
@@ -0,0 +1,137 @@
...
+        include: ${{ fromJson(needs.generate-wheels-matrix.outputs.include) }}
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up QEMU
+        if: runner.os == 'Linux'
+        uses: docker/setup-qemu-action@v2
+        with:
+          platforms: all
```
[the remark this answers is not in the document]
padix-key: OK. Could you add this explanation as comment above this step?

padix-key:
> The wheel references external versioned symbols in these
system-provided shared libraries: libc.so.6 with versions
{'GLIBC_2.2.5', 'GLIBC_2.14'}

I checked the license conditions and dynamic linking is OK for *glibc* :+1:.

> will see if there's a way to install mdtraj for tests - however it is worth mentioning that the mdtraj project is no longer being actively developed.

For this reason @JHKru currently works on replacing the *mdtraj* dependency.
I think putting these special tests into a new job, as you did, is a good solution for now. Why is it necessary to create temporary directories for these tests? The tests should not create any files.

.github/workflows/test_and_deploy.yml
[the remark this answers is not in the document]
padix-key: I wonder if this difference matter in this case, especially as the *cibuildwheel* docs also give usage examples without the action. Along with the lower complexity this would resolve the issue with synced version number, which gives me a little bit of a headache.

padix-key:
> I want to mention that the reason we need to specify 'cibuildwheel' versions in two places is that the building+tests took so long on other platforms that the time limit for free github actions was hit. Because it took so long I had to use the 'create a matrix separately' method. If the build & test times can be decreased then we won't need that step.

I will have a look to improve the runtime of the tests.

danpf:
> why is it necessary to create temporary directories for these tests? The tests should not create any files.

old habit to make sure the current directory isn't in sys.path. but it's unnecessary here.

> May I commit some changes on the test files directly to this PR?

Sure!

danpf:
Unfortunately -- debugging the time the tests take will be difficult as they are slowest on the non-native aarch64 tests.

padix-key:
I finally finished my part of the changes to the PR. I also added an action for the upload to GitHub releases in addition to PyPI. I was unable to significantly cut the test time for `AARCH46`, as it seems the runtime for all tests and also the build itself takes extremely long, presumably due to the emulation. Therefore I skipped the tests for `AARCH64`. 

As also the build takes multiple times longer than on the other platforms, I would like to propose ditching the `AARCH64` build, as to my limited knowledge it is quite a uncommon platform. The user could still install *Biotite* on `AARCH64` by installing the source distribution, although it will take longer than installing a precompiled wheel. Do we have a reason to include `AARCH64` nevertheless?

danpf:
> also added an action for the upload to GitHub releases in addition to PyPI

That's great!

> would like to propose ditching the AARCH64 build

for linux - sure! I think that's probably the best path forward. I guess I was just trying to be nice to aarch64 by keeping it in, but I don't know anyone who uses linux aarch64 anyway so it's probably just my red herring.

danpf:
Is there a reason why you made this change?
```
  CIBW_BUILD: cp39-* cp310-* cp311-*
```

Other than that - your changes look good to me!

padix-key:
> Is there a reason why you made this change?
> 
> ```
>   CIBW_BUILD: cp39-* cp310-* cp311-*
> ```
> 
> Other than that - your changes look good to me!

Up to now the official builds only supported the three most recent Python versions, which is reflected by this change. The advantage of this is the smaller build matrix and in consequence potentially smaller waiting times until GHA starts running a job.

padix-key:
> > would like to propose ditching the AARCH64 build
> 
> for linux - sure! I think that's probably the best path forward. I guess I was just trying to be nice to aarch64 by keeping it in, but I don't know anyone who uses linux aarch64 anyway so it's probably just my red herring.

I removed the `AARCH64` build. From my side the PR is ready to be merged. Thanks again for making the effort! The CI really needed refactoring.

danpf:
> Up to now the official builds only supported the three most recent Python versions

Works for me! 🎈

Issue #477, opened 2023-06-16
No way to force the sequence type when loading a FASTA file


gezmi:
Hi,

I am trying to create sequence logos from fasta files, and there I some that I cannot process but for no obvious reasons.

My code is:
```
import biotite.sequence.io.fasta as fasta
import biotite.sequence as seq

file = fasta.FastaFile()
file.read('1MFG.fa')

seqs = seq.io.fasta.get_alignment(file)
seq.SequenceProfile.from_alignment(seqs)
```

For the 1MFG.fa file, this gives
`ValueError: There is no common alphabet that extends all alphabets`
[1MFG.txt](https://github.com/biotite-dev/biotite/files/11770534/1MFG.txt)
[1JWG.txt](https://github.com/biotite-dev/biotite/files/11770538/1JWG.txt)
(I needed to convert them to txt just for uploading to github, they were named 1MFG.fa and 1JWG.fa, respectively.)

But for the 1JWG.fa file, it works. Can you help me debugging this? Both only contains the 20 common amino acids.

Thank you for you help!

padix-key:
Hi, thanks for reporting. The problem is that the sequences are very short, so that the sequence type determination of the convenience function `get_alignment()` identifies some of the sequences as nucleotide sequences. While Biotite allows `Alignment` objects of mixed sequence types (reasonable for example for protein sequence to structural alphabet alignment), this does not work for `SequenceProfile` creation. The solution would be to avoid using the convenience function and be explicit about the sequence type:
```python
import biotite.sequence.io.fasta as fasta
import biotite.sequence as seq
import biotite.sequence.align as align


def read_alignment(fasta_file):
    seq_strings = list(fasta_file.values())
    sequences = [
        # Explicit creation of ProteinSequence
        seq.ProteinSequence(seq_str.replace("-",""))
        for seq_str in seq_strings
    ]
    trace = align.Alignment.trace_from_strings(seq_strings)
    return align.Alignment(sequences, trace)

file = fasta.FastaFile()
file.read('1MFG.fa')

alignment = read_alignment(file)
print(alignment)
seq.SequenceProfile.from_alignment(alignment)
```

Although it is quite nonintuitive that `get_alignment()` may return an alignment with mixed sequence types, finding the right sequence type over a large range of sequences would probably generally decrease the performance. So in my opinion we should just mention in the documentation, that this scenario may happen.

gezmi:
[...]

PR #478, opened 2023-06-18
Add manual seq type

t0mdavid-m:
This PR implements the `seq_type` parameter as described in #477.

### DIFF
--- a/src/biotite/sequence/io/fasta/convert.py
+++ b/src/biotite/sequence/io/fasta/convert.py
@@ -220,7 +220,7 @@ def _convert_to_sequence(seq_str, seq_type=None):
                     "ProteinSequence objects do not support selenocysteine "
                     "(U), occurrences were substituted by cysteine (C)"
                 )
-            seq_str.replace("U", "C")
+            seq_str = seq_str.replace("U", "C")
         # Return the converted sequence
         return seq_type(seq_str)    
 
--- a/src/biotite/sequence/io/fasta/convert.py
+++ b/src/biotite/sequence/io/fasta/convert.py
@@ -214,12 +214,17 @@ def set_alignment(fasta_file, alignment, seq_names):
 def _convert_to_sequence(seq_str, seq_type=None):
 
     # Define preprocessing of preimplemented sequence types
+
+    # Selenocysteine is not supported, replace with cysteine
     process_protein_sequence = lambda x : x.upper().replace("U", "C")
+    # For nucleotides uracil is represented by thymine and there is is only 
+    # one letter for completely unknown nucleotides
     process_nucleotide_sequence = (
         lambda x : x.upper().replace("U","T").replace("X","N")
     )
 
-    # Attempt to set manually selected sequence type
+    # Set manually selected sequence type
+
     if seq_type is not None:
         # Do preprocessing as done without manual selection
         if seq_type == NucleotideSequence:
@@ -234,10 +239,9 @@ def _convert_to_sequence(seq_str, seq_type=None):
         # Return the converted sequence
         return seq_type(seq_str)    
 
-    # Biotite alphabets for nucleotide and proteins
+    # Attempt to automatically determine sequence type
+
     try:
-        # For nucleotides uracil is represented by thymine and
-        # there is is only one letter for completely unknown nucleotides
         return NucleotideSequence(process_nucleotide_sequence(seq_str))
     except AlphabetError:
         pass
--- a/src/biotite/sequence/io/fasta/convert.py
+++ b/src/biotite/sequence/io/fasta/convert.py
@@ -16,7 +16,7 @@ __all__ = ["get_sequence", "get_sequences", "set_sequence", "set_sequences",
            "get_alignment", "set_alignment"]
 
 
-def get_sequence(fasta_file, header=None):
+def get_sequence(fasta_file, seq_type=None, header=None):
     """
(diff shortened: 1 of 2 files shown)


src/biotite/sequence/io/fasta/convert.py
```diff
@@ -16,7 +16,7 @@
            "get_alignment", "set_alignment"]
 
 
-def get_sequence(fasta_file, header=None):
+def get_sequence(fasta_file, seq_type=None, header=None):
```
padix-key, line 19: I think for backwards compatibility reasons the new parameter should be the last parameter in the list, e.g. in case someone uses the `header` as positional argument in their code. Although Biotite is still in `0.x`version, where compatibility is not strictly required, I would prefer the *more compatible way*, if there is no clear advantage of the other option.

src/biotite/sequence/io/fasta/convert.py
```diff
@@ -28,6 +28,9 @@ def get_sequence(fasta_file, header=None):
     ----------
     fasta_file : FastaFile
         The :class:`FastaFile` to be accessed.
+    seq_type : Class, optional
+        The :class:`Sequence` subclass contained in the file. If not 
+        set, it is automatically detected.
```
padix-key, line 33: Could you also mention that only nucleotide and protein sequences can be automatically detected?

src/biotite/sequence/io/fasta/convert.py
```diff
@@ -199,26 +208,41 @@ def set_alignment(fasta_file, alignment, seq_names):
...
+            if "U" in seq_str:
+                warnings.warn(
+                    "ProteinSequence objects do not support selenocysteine "
+                    "(U), occurrences were substituted by cysteine (C)"
+                )
+            seq_str = process_protein_sequence(seq_str)
+        # Return the converted sequence
+        return seq_type(seq_str)    
+
     # Biotite alphabets for nucleotide and proteins
```
padix-key, line 234: This part of the comment is also obsolete with the code changes.

```suggestion
```

tests/sequence/test_fasta.py
```diff
@@ -40,22 +40,28 @@ def test_access_low_level():
...
         "third dna sequence" : seq.NucleotideSequence("ACGT", False),
         "rna sequence" : seq.NucleotideSequence("ACGT", False),
         "ambiguous rna sequence" : seq.NucleotideSequence("ACGTNN", True),
     }
+    path = os.path.join(data_dir("sequence"), "nuc.fasta")
+    file = fasta.FastaFile.read(path)
+    sequences = fasta.get_sequences(file)
+    assert sequences == sequences_reference
+    sequences_manual = fasta.get_sequences(file, seq.NucleotideSequence)
+    assert sequences_manual == sequences_reference
```
padix-key, line 55: This this is almost a duplicate of the two lines above, I would suggest using `pytest.mark.parametrize()`to add one version of the test with and one version without manual sequence type selection.

tests/sequence/test_fasta.py
```diff
@@ -40,22 +40,28 @@ def test_access_low_level():
...
+    assert sequences_manual == sequences_reference    
 
 
 def test_sequence_conversion():
     path = os.path.join(data_dir("sequence"), "nuc.fasta")
     file = fasta.FastaFile.read(path)
     assert seq.NucleotideSequence("ACGCTACGT") == fasta.get_sequence(file)
+    assert seq.NucleotideSequence("ACGCTACGT") == fasta.get_sequence(
+        file, seq.NucleotideSequence
+    )
```
padix-key, line 64: Same as above

tests/sequence/test_fasta.py
```diff
@@ -65,6 +71,11 @@ def test_sequence_conversion():
     # now guessed as protein sequence
     for seq1, seq2 in zip(seq_dict.values(), seq_dict2.values()):
         assert str(seq1) == str(seq2)
+    # Manually set sequence type
+    seq_dict2_manual = fasta.get_sequences(
+        file2, seq_type=seq.NucleotideSequence
+    )
+    assert seq_dict == seq_dict2_manual
```
padix-key, line 78: Same as above

padix-key (commented):
Thank you, that was quite quick! Overall your solution looks very good to me, I would only like to request some minor changes. What do you think?

padix-key:
I think we can ignore the error in `biotite.database`for this PR, there seems to be some change in the RCSB and PubChem database, that can be addressed another time.

src/biotite/sequence/io/fasta/convert.py
t0mdavid-m: I went over the comments and removed all remainders of the previous implementation. I also added/moved a comment describing the purpose of each lambda function.

src/biotite/sequence/io/fasta/convert.py
t0mdavid-m: I placed the arguments based on gut feeling. However, I agree that compatibility should take priority here.

tests/sequence/test_fasta.py
t0mdavid-m: I also refactored the test by splitting it into three functionally different tests:

- Ambiguous Sequences (i.e. nucleotide sequences which can also be loaded as amino acid sequences)
- Protein Sequences (should fail when loaded as nucleotide sequences)
- Invalid Sequences (should fail in all cases)

PR #492, opened 2023-08-21
Add fetching from the AlphaFold structure database

dacarlin:
This PR adds the ability to fetch databases from the AlphaFold structure database by querying on UniProt accession

- new functionality in `biotite.database.alphafold` module for fetching structures via UniProt accession 
- new tests in `tests/database/test_alphafold.py` 
- [to do] documentation in the user guide 
- [to do] example in the examples folder

### DIFF
--- a/src/biotite/database/alphafold/download.py
+++ b/src/biotite/database/alphafold/download.py
@@ -6,19 +6,18 @@ __name__ = "biotite.database.alphafold"
 __author__ = "Alex Carlin"
 __all__ = ["fetch"]
 
-from os.path import isdir, isfile, join, getsize
-import os
 import io
+import os
+from os.path import getsize, isdir, isfile, join
 import requests
-from .check import assert_valid_response
-
+from biotite.database.alphafold.check import assert_valid_response
 
 _fetch_url = "https://alphafold.com/api/prediction"
 
 
 def fetch(ids, target_path=None, format="pdb", overwrite=False, verbose=False):
     """
-    Download predicted protein structures from the AlphaFold DB. 
+    Download predicted protein structures from the AlphaFold DB.
 
     This function requires an internet connection.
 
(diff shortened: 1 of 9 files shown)


.github/workflows/python-package-conda.yml
```diff
@@ -0,0 +1,34 @@
+name: Python Package using Conda
```
padix-key, line 1: I think this part of the PR is out of scope. Furthermore, testing a conda build is already done in the Biotite feedstock in conda-forge. I am in favor of flake8 formatting but to me it is another topic, especially I would prefer to keep it strict and require proper flake8 formatting in all source files. Hence, this would require some effort to fix the flake8 findings in all source files.

src/biotite/database/alphafold/download.py
```diff
@@ -0,0 +1,121 @@
...
+import os
+import io
+import requests
+from .check import assert_valid_response
+
+
+_fetch_url = "https://alphafold.com/api/prediction"
+
+
+def fetch(ids, target_path=None, format="pdb", overwrite=False, verbose=False):
```
padix-key, line 19: This purely optional, but I think it would be nice to directly support the Alphafold identifiers as well. Furthermore, accepting the Alphafold identifiers given from the PDB computational models would even enable synergy between both, the `rcsb` and `alphafold` interfaces.

While using official Alphafold identifers should be straight forward (only the `v4` would need to be updated when a new version appears), using the identifier from the PDB would require some string editing. Taken from the [PDB documentation](https://www.rcsb.org/docs/general-help/computed-structure-models-and-rcsborg):

> Each CSM is assigned a specific ID in its source database and a prefix indicates the source of the model (“AF” for [AlphaFold DB](https://alphafold.ebi.ac.uk/), "MA" for [ModelArchive](https://www.modelarchive.org/)). AlphaFold DB identifiers are then followed by the UniProt accession number for the protein and by the fragment number (usually “F1”). However, in order to enable compatibility of the IDs with many of our services, including all of our APIs and visualization tools, we identify CSMs on RCSB.org using a modified version of the ID. This ID is used on the structure summary page, in searching for structures, in the search results page, and in various tools for 3D structure visualization and analysis. For example, for the AlphaFold structure AF-B3EWR1-F1, the RCSB.org assigned CSM ID is AF_AFB3EWR1F1 and is used in the query results page as shown in Figure 4.

Which type of identifier is present can be detected from whether it starts with `AF_` (PDB), `AF-` (Alphafold DB) or something else (Uniprot).

I would let you decide if we want this feature directly from the beginning

src/biotite/database/alphafold/download.py
```diff
@@ -0,0 +1,121 @@
...
+    # If input was a single ID, return only a single path
+    if single_element:
+        return files[0]
+    else:
+        return files
+
+
+
+
+
```
padix-key, line 121:
```suggestion
```

tests/database/test_alphafold.py
```diff
@@ -0,0 +1,70 @@
...
+                pdb_file = pdb.PDBFile.read(file)
+                structure = pdb_file.get_structure()
+                assert len(structure) > 0 
+            elif format == "cif":
+                cif_file = pdbx.PDBxFile.read(file) 
+                assert "citation_author" in cif_file.keys()
+
+
+
+
```
padix-key, line 70:
```suggestion
```

padix-key (changes requested):
Hi, thanks for implementing this long sought feature request. I know this is only a draft, but I had some time for review, so I can already share some thoughts :smiley:.

padix-key:
Hi, are you planning to continue work on this PR? If not, I could finish the AFDB interface.

dacarlin:
Would be amazing if you’d finish it up @padix-key

padix-key (approved):

Issue #491, opened 2023-08-21
struc.rdf doesn't work with custom box


entropybit:
The following code example where I try to caldulate the RDF with a custom box

```python
import biotite.structure as struc
import biotite.structure.io as strucio
import numpy as np

prot = strucio.load_structure("some_protein.pdb")
prot_ca = prot[prot.atom_name == "CA"]
r_com = struc.mass_center(prot_ca)
d=np.linalg.norm(prot_ca.coord - r_com.transpose(), axis=1)
#print(d.max())
box = np.array(
    [
        [2*d.max(), 0, 0],
        [0, 2*d.max(), 0],
        [0, 0, 2*d.max()]
    ]
)
prot_ca.coord = prot_ca.coord - r_com.transpose()
bins, g_r = struc.rdf(prot_ca, prot_ca, bins=10, box=box)
```

throws 
```
ValueError                                Traceback (most recent call last)
Cell In[1], line 18
     10 box = np.array(
     11     [
     12         [2*d.max(), 0, 0],
   (...)
     15     ]
     16 )
     17 prot_ca.coord = prot_ca.coord - r_com.transpose()
---> 18 bins, g_r = struc.rdf(prot_ca, prot_ca, bins=10, box=box)

File ~/miniconda3/envs/compbio/lib/python3.10/site-packages/biotite/structure/rdf.py:184, in rdf(center, atoms, selection, interval, bins, box, periodic)
    181     center = center.reshape((1,) + center.shape)
    183 if box.shape[0] != center.shape[0] or box.shape[0] != atom_coord.shape[0]:
--> 184     raise ValueError(
    185         "Center, box, and atoms must have the same model count"
    186     )
    188 # Calculate distance histogram
    189 edges = _calculate_edges(interval, bins)

ValueError: Center, box, and atoms must have the same model count
```

JHKru:
Hi, your example works for me for with (1, 3, 3)-shaped boxes.
In rdf.py, input AtomArrays are seemingly always converted to AtomArrayStacks, but the shape of (3,3)-boxes passed as arguments to the function is seemingly not adjusted to match that.

Maybe 
```
    if box is None:
        if atoms.box is None:
            raise ValueError("A box must be supplied")
        else:
            box = atoms.box
```
could be changed to something like
```
    if box is None:
        if atoms.box is None:
            raise ValueError("A box must be supplied")
        else:
            box = atoms.box
    elif box.shape == (3, 3) and atoms.stack_depth() == 1:
[...]

PR #494, opened 2023-08-28
RDF: AtomArray + (3,3)-shaped box as arguments.

JHKru:
Small fix for #491.

### DIFF
--- a/tests/structure/test_rdf.py
+++ b/tests/structure/test_rdf.py
@@ -165,7 +165,6 @@ def test_rdf_box():
         rdf(stack[:, 0], stack, box=box)
 
     # Reshape (3,3) boxes to (1,3,3) to match stacked input AtomArrays
-    print(type(stack[0]))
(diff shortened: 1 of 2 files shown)


tests/structure/test_rdf.py
```diff
@@ -145,22 +145,28 @@ def test_rdf_box():
...
     with pytest.raises(ValueError):
         rdf(stack[:, 0], stack)
 
     # Fail if box is of wrong size
     with pytest.raises(ValueError):
-        rdf(stack[:, 0], stack, box=box[0])
+        rdf(stack[:, 0], stack, box=box)
+
+    # Reshape (3,3) boxes to (1,3,3) to match stacked input AtomArrays
+    print(type(stack[0]))
```
padix-key, line 168:
```suggestion
```

padix-key (changes requested):
Thanks for the quick fix. There is only a remaining print statement, but otherwise the PR is ready to be merged.

PR #498, opened 2023-10-13
Allow 5 characters residune names.

xvlaurent:
Hello !

As you probably know, RCSB will soon be out of 3 characters codes to attribute to new ligands and plan to go for 5 characters codes before the end of the year (https://www.rcsb.org/news/65007774d78e004e766a969d).

This is an attempt of minimal modification to make Biotite compatible with such ligand codes.


padix-key:
Thanks for mentioning this, I was actually not aware of this change yet. Probably there changes on a larger scale necessary (e.g. an exception when writing PDB with 5-character residues), but this PR is already a good start. I will open an issue to tackle also the other places in the code base.

padix-key:
As I had no push access to this branch, I created PR #512. In included you commit in the new PR.

PR #541, opened 2024-03-22
Add job to periodically update the CCD

padix-key:
Closes #530.

Issue #530, opened 2024-02-03
Add CI job for updating `structure.info`


padix-key:
The *Chemical Component Dictionary* (CCD) is heavily used inside `biotite.structure.info`, e.g. for finding bonds in an `AtomArray`. The data in `biotite.structure.info` is updated using functionalities from https://github.com/biotite-dev/biotite-util.

The CCD is regularly updated by the PDB to reflect new ligands etc. However, biotite.structure.info` is not updated as regularly. It would be nice to have a regularly running CI job that creates a PR with updated information from the CCD.

padix-key:
As explained in the linked PRs, no approach to implement this issue, that brings more benefits than problems, was found. Hence, the CCD subset in version control will still be manually updated. However via #540 the dataset size is now much smaller and `structure.info` is structured more cleanly.

PR #542, opened 2024-03-24
CCD build on install

padix-key:
This PR proposes an alternative solution to #530. Instead of periodically updating the CCD dataset in version control (*Git LFS*) as implemented by #541, this PR removes the CCD dataset from version control again and instead creates it directly on installation.

Using *Git LFS* has multiple disadvanatges: First, Biotite would be locked-in into third-party version control features and, secondly, *GitHub* provides only a limited download quota from LFS which fills very fast due to CI jobs.
Furthermore, CCD creation on installation automatically ensures that the CCD version is always up-to-date.


padix-key:
This approach leads to problems in cross-compilation: `setup_ccd.py` requires importing compiled Biotite modules via `pyximport`. However in cross-compilation, these modules are compiled for the target architecture and not for the current one. Hence, an error appears. Currently I see no clean way to circumvent such problems.

PR #550, opened 2024-04-05
Migrate remaining content of `biotite-util` repository to `biotite`

padix-key:
Most functionalities from https://github.com/biotite-dev/biotite-util have been already integrated in one way or another into this repository. This PR finishes the migration. The analogs of the directories in `biotite-util` are the following:

| `biotite-util`          | `biotite`                                                 |
| ----------------------- | --------------------------------------------------------- |
| `build_gallery`         | CI job                                                    |
| `datasets`              | Refactored implementation in `setup_ccd.py`               |
| `deploy`                | CI job                                                    |
| `example_trajectories`  | Needed to be used only once (not migrated)                |
| `leontis-westhof-tests` | Migration in this PR                                      |
| `python_roles`          | Needed to be used only once (not migrated)                |
| `test_structures`       | Migration in this PR                                      |

~As the NDB has been superseded by the [NAKB](https://www.nakb.org/), which uses as different CSV format for describing base pairs (see for example https://www.nakb.org/naparams.html?1GID_1), the script cannot be used anymore to create new test data. However, probably new test data will also not be necessary in the future, so an adaption of the script is not planned.~


padix-key:
@t0mdavid-m As you are the author of the `leontis-westhof-tests`, do you agree with me that they are not migrated?

t0mdavid-m:
With minor adjustments, I created a version of the `leontis-westhof-tests` scripts that work with the NAKB database. Thus, we could create a CI-job to automatically update the test cases.

However, I am not sure if that would actually be useful as the glycosidic bond orientations and interacting edges should not change over time. In my opinion, including the generating scripts into `biotite-util` was more about providing transparency how the test data was obtained.

padix-key:
Thank you very much! I agree that the bonds will not change anymore, so we can omit a CI job for now. But it is nice to have a way to generate these files, if one would like to add entries to the test dataset in the future.

Issue #554, opened 2024-04-17
DeserializationError: Failed to deserialize category 'pdbx_struct_assembly_gen'


0ut0fcontrol:
parsing pdb nextgen archive raise DeserializationError.

example: 8h2u
<img width="1706" alt="image" src="https://github.com/biotite-dev/biotite/assets/11703338/a87e7943-07c6-4c70-bdc6-bafbf6d632a1">

rcsb cif: https://files.rcsb.org/download/8h2u.cif
nextgen cif：https://files-nextgen.wwpdb.org/pdb_nextgen/data/entries/divided/h2/pdb_00008h2u/pdb_00008h2u_xyz-enrich.cif.gz

0ut0fcontrol:
modify `_to_single()` solve the problem:

<img width="775" alt="image" src="https://github.com/biotite-dev/biotite/assets/11703338/b456ae96-e64a-431d-91a3-25ffdbcd7318">

padix-key:
Thanks for the precise pinpointing. I think this modification should work. You should even be able to replace the condition above (981 - 984) with it. As you already have the fix, would you also like to create the pull request?

PR #555, opened 2024-04-18
Fix two bugs in assembly generation

0ut0fcontrol:
Two bugs:
1. #554 
2. fix bug in parsing oper_expression '(X0)(1-10,21-25)' from 1a34


padix-key:
Looks good to me, thanks for the fixes! If you like you may add yourself to `CONTRIB.rst` as well.

We can ignore the failing doctests for now, they are caused by database changes in PubChem.

0ut0fcontrol:
I'm very glad to be a contributor. 😁

PR #557, opened 2024-04-25
Allow chain IDs with more than one character in PDB structures

GabiAxel:
Chain IDs with more than one character do exist in deposited PDB structures and therefore should be allowed in the library.

### DIFF
--- a/src/biotite/structure/io/pdb/file.py
+++ b/src/biotite/structure/io/pdb/file.py
@@ -612,7 +612,7 @@ class PDBFile(TextFile):
         if np.isnan(array.coord).any():
             raise BadStructureError("Coordinates contain 'NaN' values")
         if any([len(name) > 1 for name in array.chain_id]):
-            raise BadStructureError("Some chain IDs exceed 1 character")
+            warnings.warn("Some chain IDs exceed 1 character")
         if any([len(name) > 3 for name in array.res_name]):
             raise BadStructureError("Some residue names exceed 3 characters")
         if any([len(name) > 4 for name in array.atom_name]):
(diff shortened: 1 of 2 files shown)


padix-key:
Hi, thanks for contributing the project. In the [https://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ATOM](PDB specification) the chain ID is restricted to a single column:

> 22             Character     chainID      Chain identifier.

To my knowledge such entries exist, but are only provided in PDBx format (CIF and BinaryCIF) and not in PDB format. Do you have an example of a PDB entry where the PDB file has multi-character chain IDs?

GabiAxel:
I think you are right - in such cases the PDB format of the structure is not available, however PDB files in ECOD do exist for these chains, for example the ECOD domain e7uioAA7 includes the chain AA from 7UIO. My code failed when attempting to work with these ECOD PDB files.

padix-key:
I have to admit, I have not used ECOD before. I downloaded the mentioned entry (http://prodata.swmed.edu/ecod/complete/structure?id=e7uioAA7) and I only found the chain `A` in its `ATOM` records. So this file seems to adhere to the PDB specification. However, I also saw the chains `AA` and `BA` mentioned in the `SEQRES` records, but as I never used this database before I cannot say how the relate to the chain IDs in the `ATOM` records.

Could you add example code, that failed for you? The changes you propose would only change the behavior for writing those ECOD PDB files. This would help me understand the problem better and we also might find a solution without changing the behavior of the PDB parser.

I am quite hesitant to introduce such change: It might help working with the few use cases of PDB files that use multi-character chain IDs, deviating from the official specification. However, for the majority of cases this would allow writing PDB files that are in the *wrong* format, and hence they probably cannot be read by downstream tools.

GabiAxel:
Right again, in ECOD PDB files the chain IDs are replaces with single character IDs in these cases. The error resulted from my code attempting to rebuild these files from historic data without the chain ID replacement. Sorry about that. Closing the PR.

PR #577, opened 2024-05-31
Fix slicing with annotation on minus strand

trichter:
Hi! This is an awesome project! Very nice work.

I stumbled upon a small bug. The `sorted_locs` variable is not used in `annotation.py`. The effect is that, when slicing with a feature with at least two locations on the minus strand, we get the wrong sequence.

I added a test to demonstrate the problem and fixed it.


padix-key (approved):
Looks good to me, thanks for the fix!

PR #585, opened 2024-06-04
Stabilize test_pubchem/test_search_simple

tcjaffe:
We have an existing test against PubChem's API to retrieve ref ids.  Unfortunately, PubChem's responses will periodically change, breaking our test. This commit updates the test to handle those changes while still asserting the expected behavior of our query.


padix-key (approved):
Looks good to me. Thanks for fixing this test

PR #552, opened 2024-04-12
Revamp documentation

padix-key:
Resolves #551.

The current build of the new documentation is temporarily hosted at https://kunzmann.uber.space. Please leave a comment if you encounter any issues, typos, etc. that are not listed below yet.

**Known issues**

- [x] `News` link in menu bar misses icon
- [x] Version switcher not showing version
- [x] Search bar not working

PR #600, opened 2024-06-16
chain id parameter for structure.io.pdbx.get_sequence

tjmier:
Added a new parameter 'chain_id' to the function 'get_sequence' in the module 'biotite.structure.io.pdbx' to return a dictionary mapping chain_id to the sequence


src/biotite/structure/io/pdbx/convert.py
```diff
@@ -131,12 +131,18 @@ def get_sequence(pdbx_file, data_block=None):
...
     Returns
     -------
-    sequences : list of Sequence
-        The protein and nucleotide sequences for each entity
-        (equivalent to chains in most cases).
+    sequences : list of Sequence or dict
+        If `chain_ids` is False, returns a list of protein and nucleotide 
+        sequences for each entity.
+        If `chain_ids` is True, returns a dictionary where each key is a 
+        chain ID and each value is the corresponding sequence.
```
padix-key, line 145: I think we should note here that the chain IDs correspond to `atom_site.auth_asym_id`.

src/biotite/structure/io/pdbx/convert.py
```diff
@@ -148,7 +154,20 @@ def get_sequence(pdbx_file, data_block=None):
...
-    return sequences
+            
+    if chain_ids:
+        strand_ids = poly_category['pdbx_strand_id'].as_array(str)
+        strand_ids = [strand_id.split(",") for strand_id in strand_ids]
+
+        strand_ids_to_seq_dict = {}
+        for entity, strand_ids in enumerate(strand_ids):
+            for strand_id in strand_ids:
+                strand_ids_to_seq_dict[strand_id] = sequences[entity-1]
```
padix-key, line 165: This lines enumerates, so the first tuple value is not the entity (ID) but the index (always starting at zero).
```suggestion
        for i, strand_ids in enumerate(strand_ids):
            for strand_id in strand_ids:
                strand_ids_to_seq_dict[strand_id] = sequences[i]
```

padix-key (changes requested):
I added a few further small comments here. If we always return a dictionary, I think the function would be cleaner, if it iterates directly over strand IDs, sequence strings and sequence types instead of splitting it into two for-loops.

padix-key:
Thanks for the PR. I would be in favor of dropping the `chain_id` parameter entirely, and returning always a dictionary. Since Biotite has not reached 1.0, yet, this backwards-incompatible changes would be OK in my opinion.

Note that this would require a few small adjustments of that function call in the tests and documentation (searching for `pdbx.get_sequence` in the code base should find all occurences)

t0mdavid-m:
I agree with @padix-key. I think we should drop the `chain_id` parameter.

tjmier:
I sent an additional commit with the suggested changes as I understand them. It is worth noting that there may be instances where entity_poly.pdbx_strand_id  is not equivalent to atom_site.label_asym_id. The structure used in the test function (PDB:5UGO) is an example of this where the strand ID is "T" and the asym_id is "A". This may be an edge case and I am unsure if it will cause issues but its worth mentioning.

src/biotite/structure/io/pdbx/convert.py
```diff
@@ -134,21 +134,41 @@ def get_sequence(pdbx_file, data_block=None):
...
+    ]
+    
+    strand_ids = poly_category['pdbx_strand_id'].as_array(str)
+    strand_ids = [strand_id.split(",") for strand_id in strand_ids]
+    
+    sequence_dict = {
+        strand_id: sequence
+        for sequence, strand_ids in zip(sequences, strand_ids)
+        for strand_id in strand_ids
+    }
```
padix-key, line 169: In case some converted sequence is `None` in the list comprehension above, `sequences` would have a different length than `strand_ids`.

tjmier:
Please ignore those commits. I am still learning git and did not think those commits would be sent to this pull request. 
Again, apologies for silly mistakes. A lot of this is new to me.

padix-key:
No problem. You may also convert this PR to a Draft PR to further indicate that this is work in progress

tjmier:
The last commit is the one I would like to merge if possible.

padix-key:
Looks good to me, however, one small adjustment is necessary: With the merge of #552 `residue_coevolution.py` is moved to `doc/examples/scripts/sequence/homology/residue_coevolution.py`. This merge conflict needs to be solved first. Furthermore, you may add yourself to `CONTRIB.rst`, if you like.

PR #601, opened 2024-06-17
`Numpy 2.0` is used when installing `Biotite`via pip.

t0mdavid-m:
Currently `Numpy 2.0` is installed along `Biotite` via pip. This causes issues with imports as version 2.0 is not officially supported by `Biotite` yet. This PR fixes this issue.

@padix-key are there any other potential places in the code that we need to adjust?


padix-key:
A small but significant oversight from me. Thanks for the fix. I will also create a patch version after the merge.

PR #604, opened 2024-06-17
Fix `Atom` `__repr__()` (#602)

### LINKED ISSUE #602: Custom annotations do not appear in Atom `__repr__()`
When adding custom annotations to an `AtomArray`, they are not displayed in the `repr()` of that array. I think this is because the annotations are [hardcoded in the `__repr__` method of `Atom`](https://github.com/biotite-dev/biotite/blob/47e0a6bbe4f33f21c1a538b6bb5d955170db3b5a/src/biotite/structure/atoms.py#L499).

Showing all annotations would make it more straightforward to interactively do structural analysis, especially in a Notebook context where you frequently print atom arrays.

Below is a suggestion for a potential fix. If this would be an interesting feature to be added to Biotite I'm happy to make a PR.

```python
def __repr__(self):
    # print out key-value pairs and format strings in quotation marks
    annot_parts = [
        f'{key}="{value}"' if isinstance(value, str) else f'{key}={value}'
        for key, value in self._annot.items()
    ]

    annot = ', '.join(annot_parts)
    return f'Atom(np.{np.array_repr(self.coord)}, {annot})'
```

ljarosch:
This changes the hardcoded `Atom` `__repr__()` to a more general version that will additionally print custom added attributes.

**EDIT**: Fixes #602.

### DIFF
--- a/CONTRIB.rst
+++ b/CONTRIB.rst
@@ -4,6 +4,7 @@ CONTRIBUTORS
 - Fabrice Allain <https://github.com/f-allain>
 - Jacob Marcel Anter <https://github.com/JacobAnter>
 - Daniel Bauer <https://github.com/dnlbauer>
+- Lukas Jarosch <https://github.com/ljarosch>
 - Patrick Kunzmann <https://github.com/padix-key>
 - Tom David Müller <https://github.com/t0mdavid-m>
 - Thomas Nevolianis <https://github.com/thomasnevolianis>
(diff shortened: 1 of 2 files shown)


padix-key:
The PR looks already good to me. Would you also like to add yourself to `CONTRIB.rst`? The PR is still in draft state: Is still still work in progress or ready to be merged?

padix-key:
I linked the relevant issue in the PR description

ljarosch:
I accidentally deleted the docstring for the `__repr__` before, that's why I pulled it back to a draft. Would you prefer to have a new PR with cleaner commit history? Maybe a squash-merge would also work.

padix-key:
I will simply squash-merge.

Issue #605, opened 2024-06-22
biotite 0.41.1 `.whl` not available for MacOS ARM


BradyAJohnston:
I'm not sure on the particulars of the build pipeline - but after you tweaked the build parameters, `0.41.1` isn't available as a `.wh` for my macOS ARM machine. From looking at the versions - the ARM build is now on `macosx_10_16` where previously it was `macosx_11_0`. 

It's a niche because I'm using the pre-built `.whl` builds to package up for a Blender extension - but I'm wondering if this might be expanded to include a later `macosx` version?

padix-key:
I do not know exactly why changing the build backend changed also the wheel version, but after updating the OS version in the CI as well, the distribution is `macosx_11_0_arm64` again. We will create a patch version for it.

BradyAJohnston:
thanks mate!

padix-key:
With `0.41.2`released the correct wheel should be present. Can you confirm?

BradyAJohnston:
Yep I can download without issue now - thanks!

PR #611, opened 2024-06-24
pdbx get_sequence return dict of seq

tjmier:
Accidentally deleted my last pull request on this topic. 

**Purpose:**
Modified biotite.structure.io.pdbx.get_sequence to return a dictionary of sequences instead of a list. This change provides a more intuitive way to access sequences by their chain identifiers.

**Details:**
Old Behavior: Returned a list of sequences.
New Behavior: Returns a dictionary where keys are derived from entity_poly.pdbx_strand_id (often equivalent to chain_id and atom_site.auth_asym_id in most cases). Dictionary values are the corresponding sequences.

**Example Usage:**
```
seq_dict = get_sequences(my_cif)
chain_A_seq = seq_dict["A"]
```
**Rationale:**
The change in return type improves the usability of the function by allowing direct access to sequences using chain identifiers, which is a common use case.

**Impact:**
This change is non-backwards compatible. Code that relies on the previous list-based return type will need to be updated.
Tests and examples have been updated to reflect this change in return type.


padix-key:
For reference: This is the continuation of #600

padix-key (approved):

padix-key:
Looks good to me. I will keep this PR open until patch version 0.41.2 was released (due to #605), which should arrive in the upcoming days.

src/biotite/structure/io/pdbx/convert.py
```diff
@@ -111,7 +111,7 @@ def _filter(category, index):
 
 
 def get_sequence(pdbx_file, data_block=None):
-    """
+    """""
```
padix-key, line 114: I just wanted to merge, then I saw this typo:
```suggestion
    """
```

tjmier:
Excellent! Thank you for your patience. I learned a lot despite this being such a small change.

PR #633, opened 2024-08-08
support for YASARA PETWORLD Cif files.

corredD:
Yasara Petworld model are written in a custom cif file, that put space in front of the catergories name. 
download models from http://download.yasara.org/petworld/index.html

### DIFF
--- a/src/biotite/structure/io/pdbx/cif.py
+++ b/src/biotite/structure/io/pdbx/cif.py
@@ -465,6 +465,7 @@ class CIFCategory(_Component, MutableMapping):
         column_names = []
         i = 0
         for key_line in lines:
+            key_line = key_line.strip()
             if key_line[0] == "_":
                 # Key line
                 key = key_line.split(".")[1]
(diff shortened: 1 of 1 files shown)


BradyAJohnston:
@padix-key this is one of those examples of biotite being overly stringent when parsing that I mentioned in person last week.

padix-key:
Thanks for the fix. I checked the [CIF specification](https://www.iucr.org/resources/cif/spec/version1.1/cifsyntax) again and it seems that leading whitespaces in category names are not allowed.

> <DataItems>             ::= <Tag> <WhiteSpace> <Value>

However, in this case the additional `strip()` does increase file reading time only by approx. 3%. Hence, it would be reasonable in my opinion to support this CIF dialect.

padix-key:
@corredD If you do not have objections, I would simply apply my suggestions and merge this PR

corredD:
Sounds good to me @padix-key.

Issue #622, opened 2024-07-01
`DsspApp` is not usable with more recent `mkdssp` versions


padix-key:
As discussed in #607 `DsspApp` is not usable with more recent `mkdssp` version, limiting its usability. Hence, it should be compatible with version `3.0` (as currently) and later version. If simultaneous compatibility is not possible one could resort to a different class based on version like done for `MuscleApp`/`Muscle5App`.

padix-key:
DSSP 4 is now available on Conda https://github.com/conda-forge/staged-recipes/pull/27330

PR #637, opened 2024-08-21
Adjust Python version to make it compatible with DSSP

padix-key:
Currently the DSSP tests are disabled in `next-major-release` due to some incompatibility between *libboost* and *DSSP*. This PR aims to solve this, by temporarily decreasing the Python version in the environment again.

Having a compatible DSSP software can be tackled later, e.g. by publishing https://github.com/PDB-REDO/dssp in `conda-forge`, as *libboost* is only required for testing (https://github.com/PDB-REDO/dssp/issues/50#issuecomment-1422767236):

> The requirements are limited, only zlib development files are needed. And boost is only required for building unit-tests.

Having an updated version of DSSP available would also solve #622.

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -124,9 +124,6 @@ jobs:
         with:
           mamba-version: "*"
           environment-file: environment.yml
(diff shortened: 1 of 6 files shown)


t0mdavid-m (approved):
lgtm

PR #665, opened 2024-09-14
Add 3di encoding to `biotite.structure`

althonos:
Hi Patrick,

As discussed here is the encoder as taken (and slightly updated) from `mini3di`. 

I am not familiar with the biotite API that much but I tried doing some changes to make it fit more: I added a `StructureSequence` inheriting `Sequence` with the 3di alphabet and added support for extracting the coordinates from an `AtomArray`. This can probably be refactored a bit.


padix-key:
Thanks for providing your great code and the initial refactoring! As indicated a will further refactor it.

padix-key:
I pushed some intermediate refactored code. No worries, I am not finished yet 😉 .

padix-key:
Did you create the reference sequence in `test_3di.py` with `foldssek`? And if so, how did you tell `foldseek` just to create the 3Di sequences without aligning anything? I am asking, because it would be nice to have the 3Di sequences for the existing test structures to run the tests on them

althonos:
Hi @padix-key, you can generate the 3di sequences for arbitrary PDB files using the `makedb` and `convert2fasta` commands of foldseek, see https://github.com/steineggerlab/foldseek/issues/15.

padix-key:
Todo-list:

- [ ] Add doctests
- [ ] Add substitution matrix
- [ ] Fill missing unit tests

padix-key:
From my side the code is ready, have a look if you like. Thanks again for providing the code from your package!

One thing I do not fully understand yet, is the *invalid state* `D`. Maybe you have more insight here

- Why is it the third symbol instead of the first or last one?
- Is `D` reserved for the *invalid state* (i.e. missing atoms) or is there also a valid structure which results in `D`?

padix-key:
Ping @althonos

padix-key:
I will merge this PR now, as I plan to include this new feature in the upcoming Biotite 1.1 release. However, if you see any room for improvement or like to be added to `CONTRIB.rst`, do not hesitate to open an issue/PR.

althonos:
@padix-key : Sorry, only seeing this now!

The invalid state being 2 is actually from the original code (see https://github.com/steineggerlab/foldseek/blob/d2d09b588f50d5f8e2fd7a958377a33b2f725415/lib/3di/structureto3di.h#L9); it can also be used for valid states, as it seems to correspond to a coil state.

padix-key:
Thanks for the explanation. I also just merged your `CONTRIB.rst` addition.

PR #669, opened 2024-10-04
Allow writing out custom annotation fields to cif files

Croydon-Brixton:
feat: allow writing out custom annotation to cif file

### DIFF
--- a/tests/structure/test_pdbx.py
+++ b/tests/structure/test_pdbx.py
@@ -878,7 +878,8 @@ def test_writing_and_reading_extra_fields(tmpdir):
     # Set up a custom atom array with an additional annotation
     cif_file = pdbx.CIFFile.read(join(data_dir("structure"), "5ugo.cif"))
     atoms = pdbx.get_structure(cif_file)
-    atoms.set_annotation("my_custom_annotation", np.arange(atoms.array_length()))
+    custom_annotation = np.arange(atoms.array_length())
+    atoms.set_annotation("my_custom_annotation", custom_annotation)
 
     # Write to file
     new_cif_file = pdbx.CIFFile()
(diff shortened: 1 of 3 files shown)


padix-key:
Thanks for implementing the feature the feature! I would propose a slight change: Instead of a boolean flag to automatically include all custom annotations, I would propose an `extra_fields` parameter, where the included annotations are named, because

- it is more consistent with the `extra_fields` parameter in `get_structure()` and
- *Explicit is better than implicit* according to the Zen of Python.

Croydon-Brixton:
Great points, thank you for the feedback @padix-key -- I'll address those and update the PR

Croydon-Brixton:
✅  Alright @padix-key  - I've changed the API to use the `extra_fields` parameter as you suggested and added a mini test for the reading & writing of custom annotation.

tests/structure/test_pdbx.py
```diff
@@ -753,3 +753,29 @@ def _clear_encoding(category):
...
+
+    # Read again
+    atoms = pdbx.get_structure(
+        pdbx.CIFFile.read(join(tmpdir, "test.cif")),
+        extra_fields=["my_custom_annotation"],
+    )
+    assert "my_custom_annotation" in atoms.get_annotation_categories()
+    assert np.all(
+        atoms.get_annotation("my_custom_annotation").astype(int)
+        == np.arange(atoms.array_length())
```
padix-key, line 780: A small nitpick: I think it would be better to put `np.arange(atoms.array_length())` into an extra variable, to make clear to the reader, that this is the same as the annotation that was used in `set_annotation()`

padix-key (commented):
This looks good, I would only propose a small change, to make the test more clear.

padix-key:
I see, that you are not in `CONTRIB.rst`, yet. You are welcome to add yourself if you like.

tests/structure/test_pdbx.py
Croydon-Brixton: ✅  Done

Croydon-Brixton:
Thank you -- it'd be an honour! 
I've added myself & implemented your suggested change & rebased. Should be good to merge from my side, after CI passes -- you're welcome to squash the commits ofc

padix-key (approved):

PR #670, opened 2024-10-04
Allow loading computationally predicted cif files, which may not have certain custom annotations

Croydon-Brixton:
When loading cif files, it is convenient to specify extra fields to be loaded (`b_factor, occupancy, charge`) without having to first inspect the `atom_site` record to see whether they exist. 

This PR implements this functionality by providing fallback default values for `b_factor` (`nan`) and `occupancy` (`1.0`) in cases where no annotation is given but the extra field is nontheless required. 

It also warns when formal charge is inferred, which currently happens silently assuming a fully uncharged structure `charge=0` everywhere.

### DIFF
--- a/src/biotite/structure/io/pdbx/convert.py
+++ b/src/biotite/structure/io/pdbx/convert.py
@@ -516,7 +516,10 @@ def _fill_annotations(array, atom_site, extra_fields, use_author_fields):
     if "charge" in extra_fields:
         if "pdbx_formal_charge" in atom_site:
             array.set_annotation(
-                "charge", atom_site["pdbx_formal_charge"].as_array(int, 0)  # masked values are set to 0
+                "charge",
+                atom_site["pdbx_formal_charge"].as_array(
+                    int, 0
+                ),  # masked values are set to 0
             )
         else:
             warnings.warn(
--- a/src/biotite/structure/io/pdbx/convert.py
+++ b/src/biotite/structure/io/pdbx/convert.py
@@ -516,7 +516,7 @@ def _fill_annotations(array, atom_site, extra_fields, use_author_fields):
     if "charge" in extra_fields:
         if "pdbx_formal_charge" in atom_site:
             array.set_annotation(
-                "charge", atom_site["pdbx_formal_charge"].as_array(int)
+                "charge", atom_site["pdbx_formal_charge"].as_array(int, 0)  # masked values are set to 0
             )
         else:
             warnings.warn(
--- a/src/biotite/structure/io/pdbx/convert.py
+++ b/src/biotite/structure/io/pdbx/convert.py
@@ -480,16 +480,50 @@ def _fill_annotations(array, atom_site, extra_fields, use_author_fields):
     array.set_annotation("element", atom_site["type_symbol"].as_array(str))
 
     if "atom_id" in extra_fields:
-        array.set_annotation("atom_id", atom_site["id"].as_array(int))
+        if "id" in atom_site:
+            array.set_annotation("atom_id", atom_site["id"].as_array(int))
+        else:
+            warnings.warn(
+                "Missing 'id' in 'atom_site' category. 'atom_id' generated automatically.",
+                UserWarning,
+            )
+            array.set_annotation("atom_id", np.arange(array.array_length()))
         extra_fields.remove("atom_id")
     if "b_factor" in extra_fields:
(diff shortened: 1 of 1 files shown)


padix-key:
Thanks for the effort. However, I am somewhat divided regarding this feature.

Pro:
- Convenience, especially when scraping a lot of files
- Sensible defaults

Contra:
- Warnings can be easily overlooked, so a user might not see that the values are fabricated
- Inconsistency with the non-special extra fields, as this does not work for them

@biotite-dev Do you have opinions about this?

src/biotite/structure/io/pdbx/convert.py
```diff
@@ -480,16 +480,50 @@ def _fill_annotations(array, atom_site, extra_fields, use_author_fields):
...
+            )
+            array.set_annotation(
+                "occupancy", np.ones(array.array_length(), dtype=float)
+            )
         extra_fields.remove("occupancy")
     if "charge" in extra_fields:
-        array.set_annotation("charge", atom_site["pdbx_formal_charge"].as_array(int, 0))
+        if "pdbx_formal_charge" in atom_site:
+            array.set_annotation(
+                "charge", atom_site["pdbx_formal_charge"].as_array(int)
```
padix-key, line 519: I think there is a misunderstanding of `as_array(int, 0)`. It does not infer the charge when `"pdbx_formal_charge"` is not existing at all, but when a value in `"pdbx_formal_charge"` is masked (`.` or `?`). We still need this behavior here, as the condition above only checks if the column exists, but still charge values in many CIF files will be masked.

Actually the default value for `as_array(masked_value=None)` is still `0`, so your code still works, but in my opinion `as_array(int, 0)` is more explicit here.

nscorley:
> Thanks for the effort. However, I am somewhat divided regarding this feature.
> 
> Pro:
> 
> * Convenience, especially when scraping a lot of files
> * Sensible defaults
> 
> Contra:
> 
> * Warnings can be easily overlooked, so a user might not see that the values are fabricated
> * Inconsistency with the non-special extra fields, as this does not work for them
> 
> @biotite-dev Do you have opinions about this?

Providing some additional context why this feature may be helpful in practice: 
* During protein engineering workflows, protein designers often obtain predicted structure files in the form of PDB's or bare-bones CIF files from AlphaFold-2, RoseTTAFold All-Atom, or other deep-learning tools
* These computationally-predicted files "only" contain fully-resolved atoms and residues, and thus may not include an "occupancy" or "b-factor" column (in fact, AF-2 fills the `pLDDT` metric into the `b-factor` column)
* Ability to load these files despite not having some of RCSB's expected fields would allow Biotite to better generalize to repositories of structural data outside the Protein Data Bank

src/biotite/structure/io/pdbx/convert.py
Croydon-Brixton: Great point, thank you for clarifying and I will update ✅

JHKru:
Hi, I would also lean towards sensible defaults here.
Maybe an additional optional argument could be added for strict behaviour/errors instead of warnings, if a non-existant special field is requested with `extra_fields` (something like `strict_special_categories=False` as standard case)?
Might bloat the arguments a bit, but maybe there are use cases, where the current behaviour is beneficial.

Croydon-Brixton:
> Hi, I would also lean towards sensible defaults here. Maybe an additional optional argument could be added for strict behaviour/errors instead of warnings, if a non-existant special field is requested with `extra_fields` (something like `strict_special_categories=False` as standard case)? Might bloat the arguments a bit, but maybe there are use cases, where the current behaviour is beneficial.

That sounds like a good option to me - if you agree @padix-key , I'm happy to update the MR accordingly

padix-key:
> Maybe an additional optional argument could be added for strict behaviour/errors instead of warnings

I am inclined towards keeping the parameter list concise, especially as `get_structure()` already has quite a bunch. In the end the *strict* behavior can be enforced the same way, the user would have to go to check for available extra fields right now (without this PR): by checking for column names in `pdbx_file.block["atom_site"].keys()`.

As I have no strong opinion for or against this new behavior here and the small user base sample in this PR :wink: clearly votes for sensible defaults, I think this is a good way to go :+1:.

padix-key (approved):

padix-key:
Can you run `ruff` over the code? The code style CI check fails currently.

padix-key:
Thanks for your contribution!

Issue #673, opened 2024-10-10
Enable quoted string with whitespace in CIF file 'atom_site' category.


xvlaurent:
I have to work with in-house CIF files that have label_atom_id including whitespaces (which is allowed by the official[ mmCIF dictionary](https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_atom_site.label_atom_id.html)), but I found that biotite [does not expect whitespaces in atom_site category](https://github.com/biotite-dev/biotite/blob/231eefed334e1d3509c1b7cb3f2bfd71d4b0eeb0/src/biotite/structure/io/pdbx/cif.py#L685C1-L688C85) to optimize parsing performance.

I worked a bit on this performance issue and I can propose a function that should handle unquoting faster than the present regex implementation.
I reused the benchmark from PR #619 to compare every solutions:
```text
run number=1000000 times timeit.timeit('test_str_split(line)', number=number, globals=locals()) = 2.673100476997206 s
run number=1000000 times timeit.timeit('test_shlex_split(line)', number=number, globals=locals()) = 33.558549724999466 s
run number=1000000 times timeit.timeit('test_re_split(line)', number=number, globals=locals()) = 17.103288242000417 s
run number=1000000 times timeit.timeit('test_partitioned_split(line)', number=number, globals=locals()) = 4.747100908996799 s
```

Here is the testing code:
```python
import re
import shlex

# Declare quoting patern and pre-compile regex once
single_quote_pattern = r"('(?:'(?! )|[^'])*')(?:\s|$)"
double_quote_pattern = r'("(?:"(?! )|[^"])*")(?:\s|$)'
unquoted_pattern = r"([^\s]+)"

# Combine the patterns using alternation
combined_pattern = f"{single_quote_pattern}|{double_quote_pattern}|{unquoted_pattern}"
FINAL_REGEX = re.compile(combined_pattern)


def split_line(line):
    values = line.split()
    for k in range(len(values)):
        # Remove quotes
        if (values[k][0] == '"' and values[k][-1] == '"') or (
            values[k][0] == "'" and values[k][-1] == "'"
        ):
            values[k] = values[k][1:-1]

def _split_one_line(line):
    """
    Split a line into its fields.
    Supporting embedded quotes (' or "), like `'a dog's life'` to  `a dog's life`
    """
    # Special case of multiline value, where the line starts with ';'
    if line[0] == ";":
        return [line[1:]]
    # Find all matches
[...]

PR #686, opened 2024-10-25
Update _split_one_line and remove whitespace parameter

xvlaurent:
This is an attempt to update _split_one_line function to avoid usage of 'whitespace' parameter and approximation made for '_atom_site' CIF block.

Performance have to be approved with a CI run.

### DIFF
--- a/src/biotite/structure/io/pdbx/cif.py
+++ b/src/biotite/structure/io/pdbx/cif.py
@@ -7,7 +7,6 @@ __author__ = "Patrick Kunzmann"
 __all__ = ["CIFFile", "CIFBlock", "CIFCategory", "CIFColumn", "CIFData"]
 
 import itertools
-import re
 from collections.abc import MutableMapping, Sequence
 import numpy as np
 from biotite.file import (
@@ -1062,6 +1061,7 @@ def _split_one_line(line):
                 word, _, line = striped_line[1:].partition(separator)
 
             yield word
+
 
 def _arrayfy(data):
     if not isinstance(data, (Sequence, np.ndarray)) or isinstance(data, str):
--- a/src/biotite/structure/io/pdbx/cif.py
+++ b/src/biotite/structure/io/pdbx/cif.py
@@ -357,7 +357,7 @@ class CIFCategory(_Component, MutableMapping):
         return CIFBlock
 
     @staticmethod
-    def deserialize(text, expect_whitespace=True):
+    def deserialize(text):
         lines = [line.strip() for line in text.splitlines() if not _is_empty(line)]
 
         if _is_loop_start(lines[0]):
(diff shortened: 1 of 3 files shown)


padix-key:
Hi, thanks for writing this PR together. The benchmarks show that this variant is indeed slightly faster than than the regex-based one. However is still about 40% slower than the fast `expect_whitespace=False` hack. This makes parsing a structure without bonds about 20% slower (as shown above).

Still, I tend towards accepting this change, because it is much cleaner and technically gives the correct result, while currently it does not. If high performance is desired, I would recommend using BinaryCIF anyway. @biotite-dev has anyone a strong opinion about it?

padix-key:
Fixes #673

xvlaurent:
I tryed to reduce/optimize the code a bit more, but it seems there is no real gain...

padix-key:
This is okay, a few other people (including me) already tried that unsuccessfully. After Rust code is introduced into the project (#688), this is probably the first function that is replaced by fast Rust routine.

The PDBx test is currently failing, as it does not expect a generator. Could you either revert to the non-generator variant or adapt the test?

xvlaurent:
Finally, all tests are fine and branch is up to date!

tests/database/test_rcsb.py
```diff
@@ -31,7 +31,7 @@ def test_fetch(format, as_file_like):
     if format == "pdb":
         file = pdb.PDBFile.read(file_path_or_obj)
         pdb.get_structure(file)
-    elif format == "pdbx":
+    elif format == "cif":
```
padix-key, line 34: Good catch!

src/biotite/structure/io/pdbx/cif.py
```diff
@@ -1062,29 +1040,27 @@ def _split_one_line(line):
...
-        if field[0] == field[-1] == "'" or field[0] == field[-1] == '"':
-            field = field[1:-1]
-        fields.append(field)
-    return fields
+        yield line[1:]
+    else:
+        # Loop over the line
+        while line:
+            # Strip leading whitespace(s)
+            striped_line = line.lstrip()
```
padix-key, line 1048: Only a small typo

```suggestion
            stripped_line = line.lstrip()
```

src/biotite/structure/io/pdbx/cif.py
```diff
@@ -1062,29 +1040,27 @@ def _split_one_line(line):
...
+            # Split the line on whitespace
+            word, _, line = striped_line.partition(" ")
```
padix-key, line 1050:
```suggestion
            word, _, line = stripped_line.partition(" ")
```

src/biotite/structure/io/pdbx/cif.py
```diff
@@ -1062,29 +1040,27 @@ def _split_one_line(line):
...
+            if word.startswith(("'", '"')):
+                # Set the separator to the quote found
+                separator = word[0]
+                # Handle the case of a quoted word without space
+                if word.endswith(separator) and len(word) > 1:
+                    # Yield the word without the opening and closing quotes
+                    yield word[1:-1]
+                    continue
+                # split the word on the separator
+                word, _, line = striped_line[1:].partition(separator)
```
padix-key, line 1061:
```suggestion
                word, _, line = stripped_line[1:].partition(separator)
```

padix-key (changes requested):
I only found this small typo. Otherwise the PR looks good to merge!

Issue #689, opened 2024-10-31
Carry through space group information when reading and writing pdb files


zanebeckwith:
Thank you for the excellent project!

I noticed that, when I read in a pdb-formatted file and then write it back out,
the space group information in the `CRYST1` record gets lost.

Looking in the code, it looks like the space group is hardcoded to always be `P1`.

And, the `Z` value always gets saved as `1`.

Is there any reason the space group  and Z can't be saved when a file is read in, and properly written when the structure is saved to file?

If there is a way to do this, I'm more than happy to give it a shot and submit a PR.

Thanks!

padix-key:
`AtomArray` objects do not have a notion of space groups. Regarding the unit cell, they only store the cell vectors. Thus, `pdb.set_structure()` has no way to know the space group and only adds a placeholder space group.

In addition the support for the PDB format is relatively rudimentary in Biotite, as the format itself is deprecated, i.e. there are no low-level file editing capabilities. For PDBx (CIF/BinaryCIF) Biotite has such capabilities. You could set the space group e.g. like this:
```python
symmetry_category = CIFCategory({"space_group_name_H-M": "P 21 21 21"})
cif_file.block["symmetry"] = "P 21 21 21"
```

If you would like to implement reading/writing the space group for `PDBFile`, I would suggest to create a pair of methods `get_space_group()`, `set_space_group()`. This way the `AtomArray` does not need to to store this extra information.

zanebeckwith:
Hi @padix-key, thank you for your help!

Sorry for the long delay on my response

I think a setup like you described would work fine for our purposes. I can throw up a PR sometime soon to implement that

I have a question, though (apologies for my ignorance on crystallography):

Don't the `AtomArray` objects need to know the space group (and the `Z` value), in addition to the cell vectors?

If the pdb file contains the asymmetric unit, the space group and `Z` value will be needed to construct the full unit cell, right? Otherwise, with just the cell vectors, the `AtomArray` would have an incomplete unit cell

Again, apologies if either my understanding of crystallography is faulty here or if Biotite is already handling this gracefully

padix-key:
Implemented by #707

PR #707, opened 2024-11-27
Add support for space group information when reading and writing pdb files

alex-sbaq:
Added `get_space_group()` and `set_space_group()` methods as discussed in #689 to support space group and Z value reading/writing for (deprecated) PDB files.

### DIFF
--- a/tests/structure/io/test_pdb.py
+++ b/tests/structure/io/test_pdb.py
@@ -76,6 +76,52 @@ def test_array_conversion(path, model, hybrid36, include_bonds):
         )
     assert array1.coord.tolist() == array2.coord.tolist()
 
+@pytest.mark.parametrize(
+    "path, model",
+    itertools.product(
+        glob.glob(join(data_dir("structure"), "*.pdb")), 
+        [None, 1, -1]
+    ),
+)
+def test_space_group(path, model):
+    """
+    Test the preservation of space group information and structure 
+    when reading and writing a PDB file.
+    
+    Parameters
+    ----------
+    path : str
+        Path to the PDB file.
+    model : int or None
+        Model index for multi-model PDB files, or None to include all models.
+    """
+    # Read the PDB file
+    pdb_file = pdb.PDBFile.read(path)
+    print(f"Testing file: {path}, model: {model}")
+
+    try:
+        # Extract structure and space group
+        stack1 = pdb_file.get_structure(model=model)  # Removed duplicate argument
+        cryst1 = pdb_file.get_space_group()
+    except biotite.InvalidFileError:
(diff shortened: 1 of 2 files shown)


padix-key (changes requested):
Hi, thanks for implementing this functionality! Could you add a test as well to `tests/structure/io/test_pdb.py`? I think something analogous to `test_conversion()` would work well here: Setting and subsequent retrieving of the space group should return the original space group.

padix-key:
Thanks for the update! I will give the PR a final review on the weekend.

tests/structure/io/test_pdb.py
```diff
@@ -77,6 +77,51 @@ def test_array_conversion(path, model, hybrid36, include_bonds):
...
+        # If parsing fails for model=None due to mismatched atom counts, skip the test
+        if model is None:
+            pytest.skip("Skipping test due to incompatible atom counts across models.")
+        else:
+            raise  # Re-raise the error for other cases
+
+    # Write the structure and space group back to a new PDB file
+    pdb_file = pdb.PDBFile()
+    pdb.set_structure(pdb_file, stack1)
+    pdb.PDBFile.set_space_group(pdb_file, cryst1)
```
padix-key, line 114: I think this is more concise and idiomatic.

```suggestion
    pdb_file.set_space_group(cryst1)
```

tests/structure/io/test_pdb.py
```diff
@@ -77,6 +77,51 @@ def test_array_conversion(path, model, hybrid36, include_bonds):
     assert array1.coord.tolist() == array2.coord.tolist()
 
 
+@pytest.mark.parametrize(
+    "path, model",
+    itertools.product(glob.glob(join(data_dir("structure"), "*.pdb")), [None, 1, -1]),
```
padix-key, line 82: The number of parsed structure models does not influence the `CRYST1` record. Hence for the sake of test speed, I propose to remove the `model` parametrization from this test and always use `model=1` instead.

padix-key (changes requested):
Thanks for adding the test. I have some minor requests for it, but otherwise, everything looks good!

tests/structure/io/test_pdb.py
padix-key, line 114: I think this is more concise and idiomatic.

```suggestion
    pdb_file.set_structure(stack1)
    pdb_file.set_space_group(cryst1)
```

alex-sbaq:
Thanks for the feedback, @padix-key! I've made the changes to the test - let me know if anything is missing.

padix-key (approved):
Thanks, the PR looks good to me. I will just wait for the CI to see if the tests pass

PR #714, opened 2024-12-09
fix: allow matching `nan` values in annotations (e.g. when b_factors are set to `nan`)

Croydon-Brixton:
A tiny fix, to allow counting `nan` values as equal in annotations. This makes sense in cases where annotations are flagged as not available (e.g. b_factor = `nan`).

### DIFF
--- a/src/biotite/structure/atoms.py
+++ b/src/biotite/structure/atoms.py
@@ -255,7 +255,9 @@ class _AtomArrayBase(Copyable, metaclass=abc.ABCMeta):
         if not self.equal_annotation_categories(item):
             return False
         for name in self._annot:
-            if not np.array_equal(self._annot[name], item._annot[name], equal_nan=equal_nan):
+            if not np.array_equal(
+                self._annot[name], item._annot[name], equal_nan=equal_nan
+            ):
                 return False
         return True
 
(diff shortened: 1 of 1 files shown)


src/biotite/structure/atoms.py
```diff
@@ -233,7 +233,7 @@ def _del_element(self, index):
         else:
             raise TypeError(f"Index must be integer, not '{type(index).__name__}'")
 
-    def equal_annotations(self, item):
+    def equal_annotations(self, item, equal_nan: bool = True):
```
padix-key, line 236: Biotite does not use type annotations (yet)

```suggestion
    def equal_annotations(self, item, equal_nan=True):
```

padix-key (changes requested):
Hi, thanks for bringing this up. In my opinion you are right: When two `AtomArray`have *NaN* annotations at the same position, as a user I would expect that they are still the same. I would even suggest to be more strict and enforce this behavior (and remove the parameter), but I do not have a strong opinion here.

I only left a small nit, otherwise this looks good to me.

src/biotite/structure/atoms.py
Croydon-Brixton: Oops, thank you!

Croydon-Brixton:
Thank you for the quick review @padix-key , I've updated the docstring as you suggested & fixed ruff. Good to merge from my side.

Issue #684, opened 2024-10-23
Bonds cannot be found for alternative atom names


padix-key:
Following up #668, `chem_comp_atom.alt_atom_id` is part of the internal CCD, but is not used to determine bonds yet. However some (especially older) PDB entries contain those alternative atom names. For these `connect_via_residue_names()` cannot determine bonds, yet.

A solution would be mapping alternative atom names to their canonical ones before assigning bonds.

nscorley:
I can pick this up -- will try and open a PR within the next week

padix-key:
As mentioned in #715, this is not a problem with the latest PDB version anymore

PR #715, opened 2024-12-10
fix: support alternative atom names within connect_via_residue_names

nscorley:
Addresses https://github.com/biotite-dev/biotite/issues/684

### DIFF
--- a/src/biotite/structure/bonds.pyx
+++ b/src/biotite/structure/bonds.pyx
@@ -1647,14 +1647,16 @@ def connect_via_residue_names(atoms, bint inter_residue=True,
                 res_names[curr_start_i], {}
             )
 
-        atom_names_in_res = atom_names[curr_start_i : next_start_i]
-
         # Check if we should use alternative atom names
+        atom_names_in_res = atom_names[curr_start_i : next_start_i]
         std_atom_ids = get_from_ccd(
             "chem_comp_atom", 
             res_names[curr_start_i], 
             "atom_id"
         ) 
+
+        # Only lookup alternative atom names if we cannot match
+        # all atoms using standard names
         if (atom_names_in_res is not None and \
             std_atom_ids is not None and \
             not set(atom_names_in_res).issubset(std_atom_ids)):
@@ -1666,25 +1668,16 @@ def connect_via_residue_names(atoms, bint inter_residue=True,
                 "alt_atom_id"
             ) 
             if set(atom_names_in_res).issubset(alt_atom_ids):
-                # Standardize atom IDs
+                # Residue uses alternative names; standardize them
                 mapping = dict(zip(alt_atom_ids, std_atom_ids))
-                mapped_atom_names_in_res = np.vectorize(
-                    mapping.get
-                )(atom_names_in_res)
-                atom_names_in_res = mapped_atom_names_in_res
-
-                # If we uncomment the line below, we modify the atom_name in-place
-                # And thus enforce standardized atom names (which may be an unexpected behavior)
-                # TODO: Is that a desired behavior?
-                # atoms.atom_name[curr_start_i : next_start_i] = atom_names_in_res
-
-            # TODO: How to handle cases that do not fit either mapping?
+                atom_names_in_res = [mapping.get(atom_name) for atom_name in atom_names_in_res]
 
         for (atom_name1, atom_name2), bond_type in bond_dict_for_res.items():
             atom_indices1 = np.where(atom_names_in_res == atom_name1)[0] \
                             .astype(np.int64, copy=False)
             atom_indices2 = np.where(atom_names_in_res == atom_name2)[0] \
                             .astype(np.int64, copy=False)
+
             # In rare cases the same atom name may appear multiple times
             # (e.g. in altlocs)
             # -> create all possible bond combinations
--- a/src/biotite/structure/bonds.pyx
+++ b/src/biotite/structure/bonds.pyx
@@ -1620,6 +1620,7 @@ def connect_via_residue_names(atoms, bint inter_residue=True,
     """
     from .info.bonds import bonds_in_residue
     from .residues import get_residue_starts
+    from .info.ccd import get_from_ccd
 
     cdef list bonds = []
     cdef int res_i
@@ -1627,13 +1628,14 @@ def connect_via_residue_names(atoms, bint inter_residue=True,
     cdef int curr_start_i, next_start_i
     cdef np.ndarray atom_names = atoms.atom_name
     cdef np.ndarray atom_names_in_res
+    cdef np.ndarray std_atom_ids
     cdef np.ndarray res_names = atoms.res_name
     cdef str atom_name1, atom_name2
     cdef int64[:] atom_indices1, atom_indices2
     cdef dict bond_dict_for_res
 
     residue_starts = get_residue_starts(atoms, add_exclusive_stop=True)
-    # Omit exclsive stop in 'residue_starts'
+    # Omit exclusive stop in 'residue_starts'
     for res_i in range(len(residue_starts)-1):
         curr_start_i = residue_starts[res_i]
         next_start_i = residue_starts[res_i+1]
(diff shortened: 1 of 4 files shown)


nscorley:
@Croydon-Brixton  welcome your thoughts, as I know you've thought a lot about this as well

padix-key (changes requested):
Hi, thanks for the fix! I answered your questions above in the review comments.

padix-key:
I found a problem with residues with mixed naming standard: For about 50% of all residues in the CCD there are conflicting names between standard and alternative, i.e. the same name is used for different atoms: For example `ZZU` switches `O` and `OXT`. This means always mapping from *alt* to *std* name would mean, we accidentally identify *std* as *alt* names. Thus some bonds would not be found, even if a residue uses *std* nomenclature.

Hence, I would use your current approach: Only map atoms from a residue, if any atom name is not in the *std* list of atoms.

src/biotite/structure/bonds.pyx
```diff
@@ -1646,6 +1648,38 @@ def connect_via_residue_names(atoms, bint inter_residue=True,
             )
 
         atom_names_in_res = atom_names[curr_start_i : next_start_i]
```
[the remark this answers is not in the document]
nscorley: Good idea - will implement!

nscorley:
> I found a problem with residues with mixed naming standard: For about 50% of all residues in the CCD there are conflicting names between standard and alternative, i.e. the same name is used for different atoms: For example `ZZU` swaps `O` and `OXT`. This means always mapping from _alt_ to _std_ name would mean, we accidentally identify _std_ as _alt_ names. Thus some bonds would not be found, even if a residue uses _std_ nomenclature.
> 
> Hence, I would use your current approach: Only map atoms from a residue, if any atom name is not in the _std_ list of atoms.

Sounds like a plan. I went with the behavior to only try alternative atom names in the event we cannot map all atoms to standard names for efficiency as well — only in very rare instances (and typically only for non-canonicals/ligands) would such a mapping be needed

src/biotite/structure/bonds.pyx
```diff
@@ -1646,6 +1648,38 @@ def connect_via_residue_names(atoms, bint inter_residue=True,
...
+                # Standardize atom IDs
+                mapping = dict(zip(alt_atom_ids, std_atom_ids))
+                mapped_atom_names_in_res = np.vectorize(
+                    mapping.get
+                )(atom_names_in_res)
+                atom_names_in_res = mapped_atom_names_in_res
+
+                # If we uncomment the line below, we modify the atom_name in-place
+                # And thus enforce standardized atom names (which may be an unexpected behavior)
+                # TODO: Is that a desired behavior?
```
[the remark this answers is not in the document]
nscorley: One caveat of not updating the `AtomArray` with the mapped names is that we may not use the correct atom names when creating inter-residue bonds.

From manual inspection of the supported residues for inter-residue bonds (amino acids, DNA, RNA), this discrepancy should not be a problem — the atom names between the alternative/standard conventions are consistent.

src/biotite/structure/bonds.pyx
```diff
@@ -1646,6 +1648,38 @@ def connect_via_residue_names(atoms, bint inter_residue=True,
...
+                "chem_comp_atom", 
+                res_names[curr_start_i], 
+                "alt_atom_id"
+            ) 
+            if set(atom_names_in_res).issubset(alt_atom_ids):
+                # Standardize atom IDs
+                mapping = dict(zip(alt_atom_ids, std_atom_ids))
+                mapped_atom_names_in_res = np.vectorize(
+                    mapping.get
+                )(atom_names_in_res)
```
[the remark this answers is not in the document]
nscorley: I believe as written we only call for residues that include non-standard atom names, not every residue (since we're looping over atoms within `atom_names_in_res`); the use of `vectorize` was only for convenience. Happy to switch to the equivalent loop if preferred for readability.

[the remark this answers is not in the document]
padix-key: You are right, since we now run it for residues with non-standard atoms, performance should be a non-issue. I still slightly prefer a list comprehension for readability, but I have no strong opinion here, so you can choose.

nscorley:
Updated to address the comments; given that we only lookup alternative atom names in the rare event that we can't match all of the atoms to standard names, I thought that we didn't need a separate `get_std_to_alt_atom_name_map` function within `atoms`; happy to add if you think that would be cleaner!

src/biotite/structure/bonds.pyx
```diff
@@ -1645,12 +1647,37 @@ def connect_via_residue_names(atoms, bint inter_residue=True,
                 res_names[curr_start_i], {}
             )
 
+        # Check if we should use alternative atom names
         atom_names_in_res = atom_names[curr_start_i : next_start_i]
+        std_atom_ids = get_from_ccd(
+            "chem_comp_atom", 
+            res_names[curr_start_i], 
+            "atom_id"
+        )
```
padix-key, line 1656: `get_from_ccd()` returns a `BinaryCIFColumn` which needs to be converted into a NumPy array.

```suggestion
        std_atom_ids = get_from_ccd(
            "chem_comp_atom", 
            res_names[curr_start_i], 
            "atom_id"
        )
        std_atom_ids = std_atom_ids.as_array() if std_atom_ids is not None else None
```

src/biotite/structure/bonds.pyx
```diff
@@ -1645,12 +1647,37 @@ def connect_via_residue_names(atoms, bint inter_residue=True,
...
+            "chem_comp_atom", 
+            res_names[curr_start_i], 
+            "atom_id"
+        ) 
+
+        # Only lookup alternative atom names if we cannot match
+        # all atoms using standard names
+        if (atom_names_in_res is not None and \
+            std_atom_ids is not None and \
+            not set(atom_names_in_res).issubset(std_atom_ids)):
```
padix-key, line 1662: This line is executed for each residue, so performance is important here. Let's check if something like `np.isin(atom_names_in_res, std_atom_ids).all()` is faster, as soon as the benchmarks in the CI have run (tests need to pass first).

src/biotite/structure/bonds.pyx
```diff
@@ -1620,20 +1620,22 @@ def connect_via_residue_names(atoms, bint inter_residue=True,
...
     from .residues import get_residue_starts
+    from .info.ccd import get_from_ccd
 
     cdef list bonds = []
     cdef int res_i
     cdef int i, j
     cdef int curr_start_i, next_start_i
     cdef np.ndarray atom_names = atoms.atom_name
     cdef np.ndarray atom_names_in_res
+    cdef np.ndarray std_atom_ids
```
padix-key, line 1631: With the changes below this variable cannot be statically typed anymore as the type changes from `BinaryCIFColumn` to `ndarray`. We are not excessively accessing its methods anyway so there is probably only marginal performance benefit from typing it.

tests/structure/io/test_pdbx.py
```diff
@@ -214,6 +214,31 @@ def test_metal_coordination_bonds():
...
+    )
+    atoms_with_alt_ids = pdbx.get_structure(
+        pdbx_file_with_alt_ids, model=1, include_bonds=True
+    )
+
+    # Ensure bonds are the same
+    assert (atoms_with_alt_ids.bonds.as_array() == atoms.bonds.as_array()).all()
+
+    # Ensure atom names are the same (we mapped the alternative atom IDs to the standard atom IDs)
+    assert (atoms.atom_name == atoms_with_alt_ids.atom_name).all()
```
padix-key, line 239: This is not true, anymore, right?

tests/structure/io/test_pdbx.py
```diff
@@ -214,6 +214,31 @@ def test_metal_coordination_bonds():
...
+    )
+    atoms = pdbx.get_structure(pdbx_file_with_no_alt_ids, model=1, include_bonds=True)
+
+    # ZY9 modified to have alternative atom IDs
+    pdbx_file_with_alt_ids = pdbx.CIFFile.read(
+        join(data_dir("structure"), "6q9t_with_alt_ids.cif")
+    )
+    atoms_with_alt_ids = pdbx.get_structure(
+        pdbx_file_with_alt_ids, model=1, include_bonds=True
+    )
```
padix-key, line 233: `connect_via_residue_names()` is not specific to PDBx files, but can be used on any `AtomArray`. Hence, I propose to move such test to `test_bonds.py`. This would also remove the need to introduce a new test structure as you could run this test on `ZY9` via `structure.info.residue("ZY9")` directly, where its atom names are modified before passing it to `connect_via_residue_names()`

padix-key (changes requested):
Hi, here are a few more minor comments.

> I thought that we didn't need a separate `get_std_to_alt_atom_name_map()`

This is a good point. I would leave this decision up to you.

nscorley:
Thank you for the feedback, and apologies for the delay — I will update and re-submit this weekend!

nscorley:
Closing since in the latest updates of the PDB they no longer have outdated atom names

PR #720, opened 2024-12-15
Fix failing examples in gallery

padix-key:


padix-key:
The performance regressions are merely due to the different Python version in the CI environment

PR #730, opened 2024-12-27
fix: allow ccd residues with missing coords

davegrays:
There are a number of ligands/residues in the CCD with missing coordinates that cannot be retrieved now with `structure.info.residue` due to a ValueError that is raised twice and not excepted the second time. Relative to v0, this is new behavior.

This allows these residues to be returned from the second attempt with nans where missing.

### DIFF
--- a/src/biotite/structure/info/atoms.py
+++ b/src/biotite/structure/info/atoms.py
@@ -18,7 +18,7 @@ NON_HETERO_RESIDUES = set([
 # fmt: on
 
 
-def residue(res_name):
+def residue(res_name, allow_missing_coord=False):
     """
     Get an atom array, representing the residue with the given name.
 
@@ -30,6 +30,11 @@ def residue(res_name):
     ----------
     res_name : str
         The up to 3-letter name of the residue.
+    allow_missing_coord: bool, optional
+        Whether to allow missing coordinate values in the residue.
+        If ``True``, these will be represented as ``nan`` values.
+        If ``False``, a ``ValueError`` is raised when missing coordinates
+        are encountered.
 
     Returns
     -------
@@ -74,7 +79,11 @@ def residue(res_name):
     from biotite.structure.io.pdbx import get_component
 
     try:
-        component = get_component(get_ccd(), res_name=res_name)
+        component = get_component(
+            get_ccd(),
+            res_name=res_name,
+            allow_missing_coord=allow_missing_coord,
+        )
     except KeyError:
         raise KeyError(f"No atom information found for residue '{res_name}' in CCD")
     component.hetero[:] = res_name not in NON_HETERO_RESIDUES
--- a/src/biotite/structure/io/pdbx/convert.py
+++ b/src/biotite/structure/io/pdbx/convert.py
@@ -1299,7 +1299,7 @@ def get_component(pdbx_file, data_block=None, use_ideal_coord=True, res_name=Non
         else:
             raise
         array.coord = _parse_component_coordinates(
-            [atom_category[field] for field in alt_coord_fields]
+            [atom_category[field] for field in alt_coord_fields], keep_missing=True
         )
 
     try:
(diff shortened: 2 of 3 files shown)


padix-key:
Hi, from my perspective having a molecule without atom coordinates has barely a benefit. Consequently, getting an exception, that can be easily captured, is more clearer than checking for NaN coordinates in my opinion. However, I would be curious what your use case is.

davegrays:
@padix-key Makes sense. My use case is that I'm identifying missing atoms in the experimental structure by constructing molecular templates from the CCD (essentially just the list of atom names per residue). So for that I use the CCD's AtomArray and I'm ignoring the coords anyway.

I get your concern tho... what would you think about an optional arg for this in `get_components()` that can propagate to `structure.info.residue`?

Croydon-Brixton:
I agree with @davegrays that optionally exposing missing atoms in the builtin CCD is useful for building template residues of 'what should be there if all atoms were resolved' and thereby allows checking for missing atoms. 

I like the idea of exposing a flag in `get_components` for this -- here 's a patch MR for this https://github.com/davegrays/biotite/pull/2

davegrays:
Thanks @Croydon-Brixton ! Merged your patch

src/biotite/structure/info/atoms.py
```diff
@@ -18,7 +18,7 @@
 # fmt: on
 
 
-def residue(res_name):
+def residue(res_name, allow_missing_coords=False):
```
padix-key, line 21: Just a nit: Throughout Biotite 'coordinates' is abbreviated with `coord`, so I suggest to use it also here

```suggestion
def residue(res_name, allow_missing_coord=False):
```

src/biotite/structure/info/atoms.py
```diff
@@ -30,6 +30,11 @@ def residue(res_name):
     ----------
     res_name : str
         The up to 3-letter name of the residue.
+    allow_missing_coords: bool
```
padix-key, line 33:
```suggestion
    allow_missing_coord: bool, optional
```

src/biotite/structure/info/atoms.py
```diff
@@ -30,6 +30,11 @@ def residue(res_name):
...
+        Whether to allow missing coordinate values in the residue.
+        If `True`, these will be represented as `nan` values.
+        If `False`, a `ValueError` is raised when missing coordinates
+        are encountered.
```
padix-key, line 37: In contrast to Markdown Sphinx requires double quoted values to render them as Markdown
```suggestion
        If ``True``, these will be represented as ``nan`` values.
        If ``False``, a ``ValueError`` is raised when missing coordinates
        are encountered.
```

src/biotite/structure/io/pdbx/convert.py
```diff
@@ -1181,7 +1181,13 @@ def _filter_canonical_links(array, bond_array):
     )  # fmt: skip
 
 
-def get_component(pdbx_file, data_block=None, use_ideal_coord=True, res_name=None):
+def get_component(
+    pdbx_file,
+    data_block=None,
+    use_ideal_coord=True,
+    res_name=None,
+    allow_missing_coords=False,
```
padix-key, line 1189: See above

```suggestion
    allow_missing_coord=False,
```

src/biotite/structure/io/pdbx/convert.py
```diff
@@ -1209,6 +1215,11 @@ def get_component(pdbx_file, data_block=None, use_ideal_coord=True, res_name=Non
         In this case, the component with the given residue name is
         read.
         By default, all rows would be read in this case.
+    allow_missing_coords: bool
+        Whether to allow missing coordinate values in components.
+        If `True`, these will be represented as `nan` values.
+        If `False`, a `ValueError` is raised when missing coordinates
+        are encountered.
```
padix-key, line 1222:
```suggestion
    allow_missing_coord: bool, optional
        Whether to allow missing coordinate values in components.
        If ``True``, these will be represented as ``nan`` values.
        If ``False``, a ``ValueError`` is raised when missing coordinates
        are encountered.
```

src/biotite/structure/io/pdbx/convert.py
```diff
@@ -1330,14 +1342,20 @@ def get_component(pdbx_file, data_block=None, use_ideal_coord=True, res_name=Non
...
-def _parse_component_coordinates(coord_columns):
+def _parse_component_coordinates(coord_columns, allow_missing=False):
     coord = np.zeros((len(coord_columns[0]), 3), dtype=np.float32)
     for i, column in enumerate(coord_columns):
         if column.mask is not None and column.mask.array.any():
-            raise ValueError(
-                "Missing coordinates for some atoms",
-            )
-        coord[:, i] = column.as_array(np.float32)
+            if not allow_missing:
```
padix-key, line 1349: I would swap these blocks, i.e. `if allow_missing ... else`, to avoid the unnecessary negation.

padix-key (changes requested):
Hi, sorry for the delay. I see the point now 👍.

The PR looks mostly good to me, what misses is only a test in `tests/structure/test_info()` where `residue()` loads a molecule with missing coordinates.
Apart from that I added some small code style suggestions to the review.

davegrays:
@padix-key thanks for your feedback! all the style changes are implemented and I've added test cases for a few different molecules

davegrays:
also not sure exactly why but it seems like the ci/cd step that failed was an intermittent network-related issue

padix-key:
I just run it again, and check if it passes. But I agree, this should have nothing to do with your changes

padix-key (approved):

padix-key:
The CI now passes and everything looks good to me. Thank you!

PR #731, opened 2025-01-02
fix: expose flag to optionally parse missing cooridinates

Croydon-Brixton:


davegrays:
@Croydon-Brixton I think this is superseded now that this is merged into [here](https://github.com/biotite-dev/biotite/pull/730/) ?

Croydon-Brixton:
Oops, that was an accidental push -- thank you!

PR #732, opened 2025-01-05
feat: allow setting the `label_entity_id` from the AtomArray, if annotated

Croydon-Brixton:
The [`label_entity_id`](https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_atom_site.label_entity_id.html) is a standard category in cif files that relates the asym_id to the entity_id in entity_poly. 

Currently, when writing out an AtomArray as cif file, entity_id's are always determined anew, giving new chains new entity id's even when they are actually the same entity (try e.g. `5ocm`). This tiny PR allows to pass on the `label_entity_id` to writing out the structure. This category can already be loaded in AtomArrays via the `extra_fields` keyword argument.

### DIFF
--- a/src/biotite/structure/io/pdbx/convert.py
+++ b/src/biotite/structure/io/pdbx/convert.py
@@ -836,7 +836,11 @@ def set_structure(
     )
     atom_site["label_comp_id"] = np.copy(array.res_name)
     atom_site["label_asym_id"] = np.copy(array.chain_id)
-    atom_site["label_entity_id"] = _determine_entity_id(array.chain_id)
+    atom_site["label_entity_id"] = (
+        np.copy(array.label_entity_id)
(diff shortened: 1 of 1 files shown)


padix-key:
Hi, thanks for the PR. This seems like a sensible change to me, at least for now. In the future I would prefer to elevate `entity_id` to an optional annotation array (e.g. like `charge` or `b_factor`).

I assume this is ready to be merged?

Croydon-Brixton:
I agree, adding `entity_id` as optional annotation would be great!

And yes, ready to be merged 👍

padix-key:
Let's wait until tomorrow to see if anyone has an objection :+1:

Croydon-Brixton:
That sounds good to me!

PR #734, opened 2025-01-07
Add benchmark for `CellList`

padix-key:


Croydon-Brixton:
Hey @padix-key , I've been wondering this for a while but never had the time to investigate and thought you might know this: How does biotite's internal CellList performance compare to scipy's KDTrees (https://docs.scipy.org/doc/scipy-1.15.0/reference/generated/scipy.spatial.KDTree.html)?

The functionality they provide seem very similar (querying nearest neighbors, querying all points within a distance X, looping around in case of periodic boundary conditions, etc.) but I suspect there might have been a deeper reason to implement CellList instead of deferring to scipy's KDTrees?

padix-key:
This is a good question. There are multiple reasons:
- Years ago, when I implemented the `CellList` I did not know about `scipy.spatial.KDTree` 😉.
- The implementation is inspired by MD programs. Specifically it is able to handle periodic boundary conditions, while I do not know if K-d trees can handle them as well.
- Neighbour search in a cell list has `O(1)` time complexity while for a K-d tree it is `O(log n)` (if I am not wrong).

That being said, I would still be curious how they compare with each other in terms of performance, especially over a range of structure sizes!

Croydon-Brixton:
Thank you for the quick response - that's very helpful and makes a lot of sense!

I agree that it'd be interesting to compare the performance. I've used it on AtomArray's before when processing the entire PDB and it was very fast, but I never did a proper comparison with CellList. 

I'll put it on my backlog of things to try out -- if it were to indeed work as fast and provide the equivalent functionality, we could consider switching in KD-trees to reduce the code volume we'd need to actively maintain.

padix-key:
I generally agree. However one additional aspect to consider is that it would add `scipy` as mandatory dependency (which would probably be no big deal).

PR #741, opened 2025-01-24
Add `explicit_hydrogen` parameter

padix-key:
This parameter in `interface.rdkit.to_mol()` defines whether hydrogen should be explicitly or implicitly included in the created `Mol`

### DIFF
--- a/src/biotite/interface/rdkit/mol.py
+++ b/src/biotite/interface/rdkit/mol.py
@@ -9,16 +9,8 @@ __all__ = ["to_mol", "from_mol"]
 import warnings
 from collections import defaultdict
 import numpy as np
-from rdkit.Chem.rdchem import (
-    Atom,
-    AtomPDBResidueInfo,
-    Conformer,
-    EditableMol,
-    KekulizeException,
-    Mol,
-)
-from rdkit.Chem.rdchem import BondType as RDKitBondType
-from rdkit.Chem.rdmolops import AddHs, Kekulize, SanitizeFlags, SanitizeMol
+import rdkit.Chem.AllChem as Chem
+from rdkit.Chem import SanitizeFlags
 from biotite.interface.version import requires_version
 from biotite.interface.warning import LossyConversionWarning
 from biotite.structure.atoms import AtomArray, AtomArrayStack
@@ -31,26 +23,26 @@ _KEKULIZED_TO_AROMATIC_BOND_TYPE = {
     BondType.TRIPLE: BondType.AROMATIC_TRIPLE,
 }
 _BIOTITE_TO_RDKIT_BOND_TYPE = {
-    BondType.ANY: RDKitBondType.UNSPECIFIED,
-    BondType.SINGLE: RDKitBondType.SINGLE,
-    BondType.DOUBLE: RDKitBondType.DOUBLE,
-    BondType.TRIPLE: RDKitBondType.TRIPLE,
-    BondType.QUADRUPLE: RDKitBondType.QUADRUPLE,
-    BondType.AROMATIC_SINGLE: RDKitBondType.AROMATIC,
-    BondType.AROMATIC_DOUBLE: RDKitBondType.AROMATIC,
-    BondType.AROMATIC_TRIPLE: RDKitBondType.AROMATIC,
-    BondType.AROMATIC: RDKitBondType.AROMATIC,
+    BondType.ANY: Chem.BondType.UNSPECIFIED,
+    BondType.SINGLE: Chem.BondType.SINGLE,
+    BondType.DOUBLE: Chem.BondType.DOUBLE,
+    BondType.TRIPLE: Chem.BondType.TRIPLE,
+    BondType.QUADRUPLE: Chem.BondType.QUADRUPLE,
+    BondType.AROMATIC_SINGLE: Chem.BondType.AROMATIC,
+    BondType.AROMATIC_DOUBLE: Chem.BondType.AROMATIC,
+    BondType.AROMATIC_TRIPLE: Chem.BondType.AROMATIC,
+    BondType.AROMATIC: Chem.BondType.AROMATIC,
     # Dative bonds may lead to a KekulizeException and may potentially be deprecated
     # in the future (https://github.com/rdkit/rdkit/discussions/6995)
-    BondType.COORDINATION: RDKitBondType.SINGLE,
+    BondType.COORDINATION: Chem.BondType.SINGLE,
 }
 _RDKIT_TO_BIOTITE_BOND_TYPE = {
-    RDKitBondType.UNSPECIFIED: BondType.ANY,
-    RDKitBondType.SINGLE: BondType.SINGLE,
-    RDKitBondType.DOUBLE: BondType.DOUBLE,
-    RDKitBondType.TRIPLE: BondType.TRIPLE,
-    RDKitBondType.QUADRUPLE: BondType.QUADRUPLE,
-    RDKitBondType.DATIVE: BondType.COORDINATION,
+    Chem.BondType.UNSPECIFIED: BondType.ANY,
+    Chem.BondType.SINGLE: BondType.SINGLE,
+    Chem.BondType.DOUBLE: BondType.DOUBLE,
+    Chem.BondType.TRIPLE: BondType.TRIPLE,
+    Chem.BondType.QUADRUPLE: BondType.QUADRUPLE,
+    Chem.BondType.DATIVE: BondType.COORDINATION,
 }
 _STANDARD_ANNOTATIONS = frozenset(
     {
@@ -156,20 +148,20 @@ def to_mol(
     else:
         atoms = atoms[..., ~hydrogen_mask]
 
-    mol = EditableMol(Mol())
+    mol = Chem.EditableMol(Chem.Mol())
 
     has_annot = frozenset(atoms.get_annotation_categories())
     extra_annot = set(include_extra_annotations) - _STANDARD_ANNOTATIONS
 
     for i in range(atoms.array_length()):
-        rdkit_atom = Atom(atoms.element[i].capitalize())
+        rdkit_atom = Chem.Atom(atoms.element[i].capitalize())
         if explicit_hydrogen:
             rdkit_atom.SetNoImplicit(True)
         if "charge" in has_annot:
             rdkit_atom.SetFormalCharge(atoms.charge[i].item())
 
         # add standard pdb annotations
-        rdkit_atom_res_info = AtomPDBResidueInfo(
+        rdkit_atom_res_info = Chem.AtomPDBResidueInfo(
             atomName=atoms.atom_name[i].item(),
             residueName=atoms.res_name[i].item(),
             chainId=atoms.chain_id[i].item(),
@@ -213,7 +205,7 @@ def to_mol(
         # Handle AtomArray and AtomArrayStack consistently
         coord = coord[None, :, :]
     for model_coord in coord:
-        conformer = Conformer(mol.GetNumAtoms())
+        conformer = Chem.Conformer(mol.GetNumAtoms())
         # RDKit silently expects the data to be in C-contiguous order
         # Otherwise the coordinates would be completely misassigned
         # (https://github.com/rdkit/rdkit/issues/8221)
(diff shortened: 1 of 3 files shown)


src/biotite/interface/rdkit/mol.py
```diff
@@ -110,24 +119,29 @@ def to_mol(
...
+        atoms = atoms[..., atoms.element != "H"]
     mol = EditableMol(Mol())
 
     has_charge_annot = "charge" in atoms.get_annotation_categories()
     for i in range(atoms.array_length()):
         rdkit_atom = Atom(atoms.element[i].capitalize())
         if has_charge_annot:
             rdkit_atom.SetFormalCharge(atoms.charge[i].item())
+        if explicit_hydrogen:
+            rdkit_atom.SetNoImplicit(True)
```
Croydon-Brixton, line 132: In the base case where `explicity_hydrogen=True`, what would this mean if I just call `to_mol` for a typical crystal structure that won't have any hydrogens resolved? Will RDKit infer charges / valences in that case? If so this might lead to broken molecules -- if that were the case, should we check that if explicit hydrogen is true there **have** to be hydrogens in the structure?

padix-key: Charges would also not be inferred automatically before this PR. But I am not sure what this means for valence: As the bond types are also explicitly set I guess *RDKit* assumes a radical, if `explicit_hydrogen=True`, but hydrogen atoms are actually missing. Probably, I should test this.

Having a check there sounds like a good idea, especially as I would agree that this could be a common mistake. However, I also think strictly checking for the simple presence of hydrogen atoms might not be sensible enough, as there are valid molecules without hydrogen atoms, although they appear rarely. What do you think about raising a warning as a 'reminder' to the user to check the input?

padix-key:
I checked the different behaviors when an `AtomArray` without hydrogen is passed to `as_mol()` with `explicit_hydrogen` being `False`/`True`:

```python
import biotite.interface.rdkit as rdkit_interface
import biotite.structure.info as info
import rdkit.Chem.AllChem as Chem
import rdkit.Chem.Draw as Draw

atoms = info.residue("C")
atoms = atoms[atoms.element != "H"]

mol = rdkit_interface.to_mol(atoms, explicit_hydrogen=True)
Chem.Compute2DCoords(mol)
Draw.MolToFile(mol, "explicit.png")

mol = rdkit_interface.to_mol(atoms, explicit_hydrogen=False)
Chem.Compute2DCoords(mol)
Draw.MolToFile(mol, "non_explicit.png")
```

`explicit.png`:

explicit

`non_explicit.png`:

non_explicit

So `rdkit` correctly understands that in the latter case, that it should additional hydrogen atoms to the visualization, as the `Mol` already contains the hydrogen implicitly.

src/biotite/interface/rdkit/mol.py
```diff
@@ -141,17 +139,29 @@ def to_mol(
...
+
+    mol = Chem.EditableMol(Chem.Mol())
 
     has_annot = frozenset(atoms.get_annotation_categories())
     extra_annot = set(include_extra_annotations) - _STANDARD_ANNOTATIONS
+
     for i in range(atoms.array_length()):
-        rdkit_atom = Atom(atoms.element[i].capitalize())
+        rdkit_atom = Chem.Atom(atoms.element[i].capitalize())
+        if explicit_hydrogen:
```
Croydon-Brixton, line 158:
```suggestion
        if explicit_hydrogen:
            # ... tell RDKit to not assume any implicit hydrogens
```

Croydon-Brixton:
Sorry it wouldn't let me add this to the review summary earlier, but here's my analysis that exhibits the radicals issue:

Our base case (usual PDB files + default options)
Screenshot 2025-02-12 at 11 52 39


The other cases:
Screenshot 2025-02-12 at 11 52 52

Screenshot 2025-02-12 at 11 53 01

Screenshot 2025-02-12 at 11 53 12

Here's the code to reproduce:
```python
import biotite.structure as struc
from biotite.interface.rdkit.mol import to_mol, from_mol
import rdkit.Chem.AllChem as Chem
from typing import *

try:
    # Settings for debugging & interactive tests
    from rdkit.Chem.Draw import IPythonConsole

    IPythonConsole.kekulizeStructures = False
    IPythonConsole.drawOptions.addAtomIndices = False
    IPythonConsole.ipython_3d = False
    IPythonConsole.ipython_useSVG = True
    IPythonConsole.drawOptions.addStereoAnnotation = True
    IPythonConsole.molSize = 300, 200
except ImportError:
    pass

tyr = struc.info.residue("TYR")
tyr.chain_id[:] = "A"
tyr.res_id[:] = 17
tyr_no_h = tyr[tyr.element != "H"]

def has_radicals(mol, return_indices: bool = False) -> Union[bool, Tuple[bool, List[int]]]:
    if mol is None:
        raise ValueError("Input molecule is None")
        
    radical_indices = []
    
    for atom in mol.GetAtoms():
        # Get number of radical electrons
        num_radical_electrons = atom.GetNumRadicalElectrons()
        if num_radical_electrons > 0:
            radical_indices.append(atom.GetIdx())
    
    has_any_radicals = len(radical_indices) > 0
    
    if return_indices:
        return has_any_radicals, radical_indices
    return has_any_radicals

def is_sanitized(mol) -> bool:
    if mol is None:
        return False
        
    try:
        # Create a copy to avoid modifying the input
        mol_copy = Chem.Mol(mol)
        
        # Try to sanitize with all checks enabled
        # catchErrors=True prevents exceptions from being raised
        result = Chem.SanitizeMol(mol_copy, catchErrors=True)
        
        # SANITIZE_NONE (0) means all checks passed
        return result == Chem.SanitizeFlags.SANITIZE_NONE
        
    except Exception:
        return False
    
rad_color = lambda has_rad: '\033[32m' if not has_rad else '\033[31m'  # Green if False, Red if True
san_color = lambda is_san: '\033[32m' if is_san else '\033[31m'      # Green if True, Red if False
reset_color = '\033[0m'

i = 0
for m in [tyr_no_h, tyr]:
    for explicit_hydrogen in [True, False]:
        mol = to_mol(m, explicit_hydrogen=explicit_hydrogen)
        Chem.Compute2DCoords(mol)
        print(f"------- {i}: has_H={'H' in m.element}, {explicit_hydrogen=} -------")
        
        # Color the output based on the conditions
        print("before sanitize")
        has_rad = has_radicals(mol)
        is_san = is_sanitized(mol)
        print(f"{rad_color(has_rad)}has_radicals(mol): {has_rad}{reset_color}")
        print(f"{san_color(is_san)}is_sanitized(mol): {is_san}{reset_color}")
        display(mol);

        Chem.SanitizeMol(mol)
        print("after sanitize")
        has_rad = has_radicals(mol)
        is_san = is_sanitized(mol)
        print(f"{rad_color(has_rad)}has_radicals(mol): {has_rad}{reset_color}")
        print(f"{san_color(is_san)}is_sanitized(mol): {is_san}{reset_color}")
        display(mol);

        print("converted back into biotite structure")
        print(repr(from_mol(mol, conformer_id=0)))

        i += 1
```

Croydon-Brixton:
Not directly related to the radicals, but another thing I noted while looking at this:
The `from_mol` function returns an empty stack when no conformer is given for this test case. I think that is undesired and we probably want to fix this as well.

Screenshot 2025-02-12 at 11 55 04

Croydon-Brixton:
@padix-key I've added my suggested changes here for your convenience: https://github.com/padix-key/biotite/pull/13

src/biotite/interface/rdkit/mol.py
```diff
@@ -141,17 +139,29 @@ def to_mol(
     HB3
     HXT
     """
-    mol = EditableMol(Mol())
+    hydrogen_mask = atoms.element == "H"
```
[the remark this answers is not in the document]
padix-key: This is indeed an important edge case! I agree, that deuterium should be handled like hydrogen. However, this problem is not isolated to `to_mol()` but involves difference places in the code base. Hence I added a separate issue (#758) and propose to handle this afterwards.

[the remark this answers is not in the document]
padix-key: I also thought about it again, and I would prefer not removing hydrogen atoms, if `explicit_hydrogen=False`, but raise an exception instead. This would ensure that the atom indices of the created `Mol` always corresponds to the atom indices of the input `AtomArray`. This would for example matter when using indices obtained from `Mol.GetSubstructMatch()` to filter the `AtomArray`.

padix-key:
Thanks @Croydon-Brixton for your support. Your suggested changes look basically good to me. I would have some minor suggestions, but it is probably less confusing for you, if I bring them up in this PR.

src/biotite/interface/rdkit/mol.py
```diff
@@ -352,22 +398,40 @@ def from_mol(mol, conformer_id=None, add_hydrogen=None):
             extra_annotations[annot_name][_atom_idx] = rdkit_atom.GetProp(annot_name)
 
     for annot_name, array in extra_annotations.items():
-        atoms.set_annotation(annot_name, array.astype(str))
+        # ... handle special case of implicit hydrogen atom flags
+        if annot_name == "isImplicit":
+            atoms.set_annotation(annot_name, array.astype(bool))
+            continue
```
padix-key, line 404: Could you elaborate on this, i.e. is there an such an Atom property?

padix-key:
I mainly added two changes:

- The default value of `explicit_hydrogen` is `None`, which infers the argument from the presence of hydrogen atoms.
- I added the ability to select either 2D or 3D conformers. If no (matching) conformer is present, a model with *NaN* coordinates is returned.

src/biotite/interface/rdkit/mol.py
padix-key: Now I also encountered it :+1:. I expanded the comment a bit, explaining that `AddHs()` is the source.

padix-key:
@Croydon-Brixton Do you think this PR is ready to merge?

src/biotite/interface/rdkit/mol.py
```diff
@@ -105,6 +100,12 @@ def to_mol(
         are always included per default. These standard annotations can be accessed
         with :meth:`rdkit.Chem.rdchem.Atom.GetPDBResidueInfo()` for each atom in the
         returned :class:`rdkit.Chem.rdchem.Mol`.
+    explicit_hydrogen : bool, optional
+        If set to true, the conversion process expects that all hydrogen atoms are
+        explicit, i.e. each each hydrogen atom must be part of the :class:`AtomArray`.
+        If set to false, the conversion process treats all hydrogen atoms as implicit.
+        By default, explicit hydrogen atoms are only assumed of any hydrogen atoms are
```
Croydon-Brixton, line 107:
```suggestion
        By default, explicit hydrogen atoms are only assumed if any hydrogen atoms are
```

src/biotite/interface/rdkit/mol.py
```diff
@@ -215,9 +255,12 @@ def from_mol(mol, conformer_id=None, add_hydrogen=None):
     ----------
     mol : rdkit.Chem.rdchem.Mol
         The molecule to be converted.
-    conformer_id : int, optional
-        The conformer to be converted.
-        By default, an :class:`AtomArrayStack` with all conformers is returned.
+    conformer_id : int or {"2D", "3D"}, optional
```
Croydon-Brixton, line 258: Very nice!

tests/interface/test_rdkit.py
```diff
@@ -145,15 +163,40 @@ def test_kekulization():
...
+    ref_atoms.set_annotation(EXTRA_ANNOT_NAME, ref_annot)
+
+    mol = rdkit_interface.to_mol(
+        ref_atoms, include_extra_annotations=[EXTRA_ANNOT_NAME]
+    )
+    test_atoms = rdkit_interface.from_mol(mol)
+    test_annot = test_atoms.get_annotation(EXTRA_ANNOT_NAME)
+
+    assert test_annot.dtype == ref_annot.dtype
+    assert test_annot.tolist() == ref_annot.tolist()
```
Croydon-Brixton, line 188: Great to see the annotation dtypes preserved! Well done for finding a way to do that :D

Croydon-Brixton (approved):
Oops, sorry the requested re-review somehow slipped through my fingers!

Yes, the changes look all good to me and from my point of view this is ready to merge. While reading over it again I just found one tiny typo.

- I like the new default of 'None' for our explicit hydrogen policy. I think inferring makes a lot of sense and the edge cases where this isn't desired (e.g. some small molecule that truly has no hydrogens) should be harmless, since re-inferring should conclude that indeed no hydrogens are needed.
- The selection of 2D & 3D conformers is very neat, thank you!!
- Thank you for finding a way to keep the annotation dtypes preserved where possible!

PR #745, opened 2025-01-29
Feat/update vdw radii

Croydon-Brixton:
I had a detailed look into the latest open access vdW radii, especially for elements that were not previously represented in biotite. I updated the values with recent authoritative sources, which should now allow SASA computations for almost all structures. I also did a comparison against non-open access reference values which we cannot use here, but made sure that open access values we use are close to (usually <0.05A) those reference values.


Croydon-Brixton:
Not sure what's happening on the CI side here as I only changed a dictionary -- maybe a glitch?

src/biotite/structure/sasa.pyx
```diff
@@ -83,7 +83,8 @@ def sasa(array, float probe_radius=1.4, np.ndarray atom_filter=None,
             - **Single** - A set, which uses a defined VdW radius for
               every single atom, therefore hydrogen atoms are required
               in the model (e.g. NMR elucidated structures).
-              :footcite:`Bondi1964`
+              Values for main group elements are taken from :footcite:`Mantina2009`,
+              and for relevant transition metals from the RDKit.
```
padix-key, line 87: Could you also add a citation to the relevant RDKit paper here?

```suggestion
              and for relevant transition metals from *RDKit*.
```

src/biotite/structure/info/radii.py
```diff
@@ -26,37 +26,106 @@
...
-    ("BR", 1, 0) : 1.85, # Taken from _SINGLE_RADII
+    ("F",  1, 0) : 1.47, # Taken from _SINGLE_ELEMENT_VDW_RADII
+    ("CL", 1, 0) : 1.75, # Taken from _SINGLE_ELEMENT_VDW_RADII
+    ("BR", 1, 0) : 1.85, # Taken from _SINGLE_ELEMENT_VDW_RADII
     ("I",  1, 0) : 1.98, # Taken from _SINGLE_RADII
 }
 
-_SINGLE_RADII = {
-    "H":  1.20,
+_SINGLE_ELEMENT_VDW_RADII = {
```
padix-key, line 35: Just an optional nit: I think *single atom* is more appropriate than *single element* here.

```suggestion
_SINGLE_ATOM_VDW_RADII = {
```

padix-key (commented):
Thanks for the update. I only left two minor comments, otherwise this looks good!

I will also add the new reference (`Mantina2009`) to the Biotite Zotero library.

Regarding the failing test: It fails, because the reference SASA values were computed with `mdtraj` using the old radii. To update the reference values to the new radii, just rerun [`sasa.py`](https://github.com/biotite-dev/biotite/blob/main/tests/structure/data/misc/sasa.py) (requires `mdtraj`). I can support you here, if you need some help.

src/biotite/structure/sasa.pyx
Croydon-Brixton: Done ✅

src/biotite/structure/info/radii.py
Croydon-Brixton: Done ✅

Croydon-Brixton:
Alright, SASA values & tests are updated. Two small points, mostly recorded here for posterity:

1. The values changed quite a bit - mostly due to the 1.2A (Bondi)--> 1.1A (Mantina) change in Hydrogen. Given the primary literature, I think the new, updated value of 1.1A is more accurate with the latest scientific findings:

Here is what Mantina et al. 2009 write about it
> In spite of the various caveats that may be made, Bondi’s values for the van der Waals radii are certainly very reasonable, the best argument for which is that they have been found useful for an inordinately large number of applications by later workers. The value for hydrogen atom has, however, been found to problematic. Rowland and Taylor,[14](https://pmc.ncbi.nlm.nih.gov/articles/PMC3658832/#R14) taking advantage of the large amount of additional crystal data that has been accumulated since Bondi’s study, made a systematic analyses of intermolecular contact distances in organic crystals and redetermined the values of the van der Waals radii. They found that Bondi’s values were consistent with the new data with the single exception of hydrogen, for which they suggested lowering the van der Waals radius from 1.2 Å to 1.1 Å. Bondi’s value of 1.2 Å describes well the size of H in H•••H contacts, but for interactions with other elements, the smaller value is more appropriate. Bondi had drawn similar conclusions, but he favored using the higher value as a standard. We accept the Rowland-Taylor recommendation for H.
For what it's worth, the CRC also uses the updated value.

2. Popular software packages such as [mdtraj](https://github.com/mdtraj/mdtraj/blob/de18b88c4722776626a8818edd8ba0e96e1144cb/mdtraj/geometry/sasa.py#L33-L176) use ionic radii for the metals rather than van der Waals radii, with the explanation:

> # However, because certain atoms in biophysical simulations have a high
# chance of being completely ionized, we have decided to give the
# following atoms their ionic radii their ionic radii:
# +2: Be, Mg, Ca, Ba
# +1: Li, Na, K, Cs
# -1: Cl

I believe this is wrong and we should be using the VdW radius instead (as proposed in this PR).  Ultimately, SASA is asking the question “How close can a solvent molecule (often approximated by a water sphere) get to each atom?” This “solvent accessible” boundary is governed primarily by short‐range interatomic repulsions, not by how small the bare ionic core is once you strip away bonding electrons -- which can be well under 1A. For example, the ionic radius of Mg2+ is around 0.7A -- solvent, i.e. water does not usually ever come close this little of a distance, instead if it's e.g. involved in hydrogen bonding networks with that ion it sit more or less around the 2A ballpark (which is more in line with the VdW radii, I am proposing to use here -- for Mg it's about 1.8A).
Besides that, VdW radii have the advantage of being less dependent on the ionization state and generally easier to measure -- [coming with lower variances](https://www.chemguide.co.uk/atoms/properties/atradius.html) -- , from what I can read online.

Croydon-Brixton:
Alright, should be all good from my side now

padix-key (approved):
Thanks for the detailed explanation. I totally agree with your argumentation. An eventually, if someone does want to compute a SASA variant using ionic radii, they could simply provide the custom radii via the `vdw_radii` parameter.

Issue #709, opened 2024-11-29
Adding interfaces to popular Python packages


padix-key:
Although Biotite implements a lot of algorithms from the bioinformatics community, there are still many tasks that cannot be done in Biotite, either because a method is not implemented yet, it is too complex or because it is out of scope. For this purpose Biotite provides the `biotite.application` subpackage, which provides interfaces to external software.
The logic of the core `biotite.application.Application` class inherently assumes that another process is running the desired functionality - either local (e.g. `Muscle` )or on some server (currently only *NCBI BLAST*).

However, there are also a lot of popular Python packages for various bionformatics tasks out there, where an interface would be reasonable in my opinion. This would be a non-exhaustive list:

- [x] [PyMOL](https://github.com/schrodinger/pymol-open-source) (#749)
- [x] [OpenMM](https://github.com/openmm/openmm) (#750)
- [ ] [ViennaRNA](https://github.com/ViennaRNA/ViennaRNA) (see #708 as motivation)
- [x] [RDKit](https://github.com/rdkit/rdkit) (#736)
- [ ] [Biopython](https://github.com/biopython/biopython)
- [ ] [pyOpenMS](https://github.com/openMS/openms)

Pressing these into the logic of the `Application` would be possible, but

- it would be slower, due to file I/O instead of converting Python objects,
- it would be counterintuitive, as users probably expect to use a Python library as library instead of some static application run and
- it would be inflexible, as one would be unable to use the full extent of the programming language

Hence, I would like to discuss an additional `biotite.interface` subpackage: Analogous to `biotite.application` it would contain the `biotite.interface.xyz` subpackage for the Python package `xyz` and the user is responsible for installing `xyz` separately. However, unlike `biotite.application` its purpose is to convert between native `biotite` objects and the corresponding objects in `xyz`. So the user could use the converted object directly within the same Python script.

Pros:

- A pythonic way to interface other packages in the bioinformatics ecosystem.
- Ability to migrate some current extension packages directly into `biotite.interface` to reduce maintenance overhead (namely [`ammolite`](https://github.com/biotite-dev/ammolite) and [`molmarbles`](https://github.com/biotite-dev/molmarbles)).

Cons:

- The conversion code needs to be maintained as well and needs to be kept up to date with the interfaced package.
- Breaking changes in the interfaced package lead to ambiguity: For which version of the package does the conversion function work?
  - Could possibly solved with a function decorator that specifies the version.

[...]

PR #750, opened 2025-02-09
Integrate `openmm` into `biotite.interface.openmm`

padix-key:
As described in #709, this PR adds an interface to the *OpenMM* molecular simulation toolkit.
The code is taken and adapted from [`biotite-dev/molmarbles`](https://github.com/biotite-dev/molmarbles/). Afther this PR is merged, [`biotite-dev/molmarbles`](https://github.com/biotite-dev/molmarbles/) will be archived.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -112,7 +112,10 @@ bibtex_default_style = "ieee"
 
 notfound_urls_prefix = "/latest/"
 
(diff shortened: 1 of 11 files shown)


padix-key:
The failing test is fixed in https://github.com/biotite-dev/biotite/pull/761.

PR #751, opened 2025-02-11
Add functionalities related to aromatic rings

padix-key:
This PR adds the `find_aromatic_rings()` and `find_stacking_interactions()` functions to `biotite.structure`, that can be used to identify atoms that are part of aromatic rings and find pi-stacking interactions between them, respectively.


padix-key:
Thanks, I also stumbled upon it during my research and adapted the cutoff distance from it.

PR #762, opened 2025-02-24
Add `biotite.interface` subpackage

padix-key:
This PR adds the `biotite.interface` subpackage as discussed in #709. It ships the following interfaces:

- `biotite.interface.rdkit`
- `biotite.interface.pymol`
- `biotite.interface.openmm`

along with tutorials for each interface.

All subpackages were already part of previous PRs merging into the `interfaces` branch. This PR will bring the new interfaces collectively into the `main` branch. Let me know, if there are any requests before merging @t0mdavid-m @JacobAnter @Croydon-Brixton @MaxGreil .


Croydon-Brixton:
Very excited to have these in & continue building on them! Thank you for spearheading this @padix-key !

src/biotite/interface/rdkit/mol.py
```diff
@@ -0,0 +1,482 @@
...
+        # Handle AtomArray and AtomArrayStack consistently
+        coord = coord[None, :, :]
+    for model_coord in coord:
+        conformer = Chem.Conformer(mol.GetNumAtoms())
+        # RDKit silently expects the data to be in C-contiguous order
+        # Otherwise the coordinates would be completely misassigned
+        # (https://github.com/rdkit/rdkit/issues/8221)
+        conformer.SetPositions(np.ascontiguousarray(model_coord, dtype=np.float64))
+        conformer.Set3D(True)
+        mol.AddConformer(conformer)
```
Croydon-Brixton, line 243: @padix-key 
One more element that'd be useful to set would be the stereochemistry of the molecule from the coordinates:
`Chem.AssignStereochemistryFrom3D` (or at least make this an option). Otherwise we might end up with D-Amino acids, etc when ppl delete the original conformer and re-sample one. Something that'll probably quite unexpected & undesired for most users.

padix-key: In my opinion doing such post-conversion cleanups should be the responsibility of the user:

- Although for some applications stereochemistry is useful (another one would be substructure matching), for many applications it is not required.
- Running such cleanups requires additional computation time.
- There are many different 'handy' cleanups in RDKit (`AssignStereochemistryFrom3D`, `SanitizeMol`, `UFFOptimizeMolecule`, etc.). So running them (including having the parameters, whether they should be run) would broaden the functionality too much in my opinion. Instead `to_mol()` and `from_mol()` are currently only concerned with converting structure information between Biotite and RDKit with as few information loss as possible.
- Also running the cleanups by the user after calling `to_mol()` is relatively simple one-liner.

Croydon-Brixton: Hmm good point, agreed it'd be a one-liner and probably best to keep our API simple!

t0mdavid-m:
Sorry I am currently at the end of my vacation and missed this. I will review it next week and also submit a PR for the interface to `pyopenms`.

padix-key:
No worries. I already merged this PR, but if you request any changes, we can open a new issue for that.

padix-key:
@t0mdavid-m Did you already have the opportunity to look at the new features? I am currently planning to release Biotite `1.2` in the next few days, but I first wanted to make sure, that you are happy with the new `interface` subpackage, especially if you would request any backwards incompatible changes.

t0mdavid-m:
@padix-key I just looked at the new features. I think `to` and `from` functions for each datastructure are sensible and I really like the way compatibilities handled. Looks like this should also work with multiple constraints from the same package (i.e. `["package>1.2", "package<1.4"]`. In my opinion that is pretty future-proof. From my side you can go ahead and release `1.2`.

I will create the Interface to the `pyOpenMS` data structures soon (our developers conference is next week so there should be time during a Hackathon).

PR #765, opened 2025-03-03
Optimize memory utilization of `get_structure()`.

cloverzizi:
When obtaining an AtomArray using `pdbx.get_structure()`, the invoked `_find_matches()` function generates a `[N_struct_conn_col, N_struct_conn_row, N_atom]` matrix. In cases where a CIF file contains numerous atoms and numerous inter-residue bonds, this can lead to a significant increase in memory usage. For example, for PDB ID 7Y4L, memory consumption can reach up to 100GB, and it takes approximately 450 seconds to run `pdbx.get_structure()`.
This PR modifies the implementation of `_find_matches()` to address and prevent this issue. With the updated code, processing 7Y4L requires less than 1.5GB of memory, and the runtime is reduced to approximately 45 seconds.

### DIFF
--- a/src/biotite/structure/io/pdbx/convert.py
+++ b/src/biotite/structure/io/pdbx/convert.py
@@ -104,8 +104,8 @@ CANONICAL_RESIDUE_LIST = canonical_aa_list + canonical_nucleotide_list
 # exceed a certain threshold,
 # a dictionary approach is less computation and memory intensive than the dense
 # vectorized approach.
-# https://github.com/biotite-dev/biotite/pull/765#issuecomment-2696591338
-FIND_MATCHES_SWITCH_THRESHOLD = 10000
+# https://github.com/biotite-dev/biotite/pull/765#issuecomment-2708867357
+FIND_MATCHES_SWITCH_THRESHOLD = 4000000
 
 _proteinseq_type_list = ["polypeptide(D)", "polypeptide(L)"]
 _nucleotideseq_type_list = [
--- a/src/biotite/structure/io/pdbx/convert.py
+++ b/src/biotite/structure/io/pdbx/convert.py
@@ -115,6 +115,8 @@ _other_type_list = [
     "polysaccharide(L)",
 ]
 
+FIND_MATCHES_SWITCH_THRES = 10000
+
 
 def _filter(category, index):
     """
@@ -644,7 +646,7 @@ def _parse_inter_residue_bonds(atom_site, struct_conn):
     )
 
 
-def _find_matches(query_arrays, reference_arrays):
+def _find_matches_by_dense_array(query_arrays, reference_arrays):
     """
     For each index in the `query_arrays` find the indices in the
     `reference_arrays` where all query values match the reference counterpart.
@@ -674,6 +676,63 @@ def _find_matches(query_arrays, reference_arrays):
     # -1 indicates that no match was found in the reference
     match_indices = np.full(len(query_arrays[0]), -1, dtype=int)
     match_indices[query_matches] = reference_matches
+    return match_indices
+
+
+def _find_matches_by_dict(query_arrays, reference_arrays):
+    """
+    For each index in the `query_arrays` find the indices in the
+    `reference_arrays` where all query values match the reference counterpart.
+    If no match is found for a query, the corresponding index is -1.
+    """
+    # Convert reference arrays to a dictionary for O(1) lookups
+    reference_dict = {}
+    ambiguous_keys = set()
+    for ref_idx, ref_row in enumerate(zip(*reference_arrays)):
+        ref_key = tuple(ref_row)
+        if ref_key in reference_dict:
+            ambiguous_keys.add(ref_key)
+            continue
+        reference_dict[ref_key] = ref_idx
+
+    match_indices = []
+    for query_idx, query_row in enumerate(zip(*query_arrays)):
+        query_key = tuple(query_row)
+        occurrence = reference_dict.get(query_key)
+
+        if occurrence is None:
+            # -1 indicates that no match was found in the reference
+            match_indices.append(-1)
+        elif query_key in ambiguous_keys:
+            # The query cannot be uniquely matched to an atom in the reference
+            raise InvalidFileError(
+                f"The covalent bond in the 'struct_conn' category at index "
+                f"{query_idx} cannot be unambiguously assigned to atoms in "
+                f"the 'atom_site' category"
+            )
+        else:
+            match_indices.append(occurrence)
+
+    return np.array(match_indices)
+
+
+def _find_matches(query_arrays, reference_arrays):
+    """
+    For each index in the `query_arrays` find the indices in the
+    `reference_arrays` where all query values match the reference counterpart.
+    If no match is found for a query, the corresponding index is -1.
+    """
+    # it was observed that when the size exceeds 2^13 (8192)
+    # the dict strategy becomes significantly faster than the dense array
+    # and does not cause excessive memory usage.
+    # https://github.com/biotite-dev/biotite/pull/765#issuecomment-2696591338
+    if (
+        query_arrays[0].shape[0] * reference_arrays[0].shape[0]
+        <= FIND_MATCHES_SWITCH_THRES
+    ):
+        match_indices = _find_matches_by_dense_array(query_arrays, reference_arrays)
+    else:
+        match_indices = _find_matches_by_dict(query_arrays, reference_arrays)
     return match_indices
 
 
--- a/src/biotite/structure/io/pdbx/convert.py
+++ b/src/biotite/structure/io/pdbx/convert.py
@@ -100,6 +100,12 @@ COMP_BOND_TYPE_TO_ORDER = {
     bond_type: order for order, bond_type in COMP_BOND_ORDER_TO_TYPE.items()
 }
 CANONICAL_RESIDUE_LIST = canonical_aa_list + canonical_nucleotide_list
+# it was observed that when the number or rows in `atom_site` and `struct_conn`
+# exceed a certain threshold,
+# a dictionary approach is less computation and memory intensive than the dense
+# vectorized approach.
+# https://github.com/biotite-dev/biotite/pull/765#issuecomment-2696591338
+FIND_MATCHES_SWITCH_THRESHOLD = 10000
 
 _proteinseq_type_list = ["polypeptide(D)", "polypeptide(L)"]
 _nucleotideseq_type_list = [
@@ -114,8 +120,6 @@ _other_type_list = [
     "polysaccharide(D)",
     "polysaccharide(L)",
 ]
-
-FIND_MATCHES_SWITCH_THRES = 10000
 
 
 def _filter(category, index):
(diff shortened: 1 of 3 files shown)


Croydon-Brixton (changes requested):
Thank you for this nice addition!

This looks great to me, my only slight concern is the 20-30% performance hit we get on average-sized structures, which is the usual case for almost all PDB structures. 

One simple way to get the best of both worlds would be to pick between the two strategies (dict strategy vs dense array strategy) based on the size of 'query_arrays', 'reference_arrays' and 'n_atoms'. Could you add a switch use the array based matching for typical structures, but the dictionary based matching for problems above a certain size? This should allow us to retain the benchmark speed while fixing the issue you observed for large structures.
(Alternatively you could also implement the dictionary matching in cython, I'd imagine that should also fix the slowdown).

Also, would you be able to add your observed test case to the test suite and mark it as slow test (via pytest marks)? This way we can track the performance on large structures as well.

src/biotite/structure/io/pdbx/convert.py
```diff
@@ -650,31 +650,35 @@ def _find_matches(query_arrays, reference_arrays):
...
-        raise InvalidFileError(
-            f"The covalent bond in the 'struct_conn' category at index "
-            f"{ambiguous_query} cannot be unambiguously assigned to atoms in "
-            f"the 'atom_site' category"
-        )
+    # Convert reference arrays to a dictionary for O(1) lookups
+    reference_dict = {}
+    unambiguously_keys = set()
+    for idx, col in enumerate(np.stack(reference_arrays, axis=-1)):
+        ref_key = "_".join(map(str, col))
```
padix-key, line 657: Is there a reason to convert specifically to string here? I assume we could save computation time if we used just a tuple of the original values.

cloverzizi:
@Croydon-Brixton @padix-key Thank you both for your suggestions. I initially tried using tuples as dictionary keys, which improved performance somewhat, but it still did not outperform the dense array strategy in the benchmark test (1AKI). As a result, I implemented a switch based on the size of N_query * N_ref, ultimately choosing 2^3 as the threshold to decide which strategy to use.

Additionally, I did not include 7Y4L in the pytest testing code because its CIF file is excessively large (>100MB), which would bloat the repository size.

image
image

padix-key:
Thanks for the helpful benchmark! I would not have expected that the tipping point comes so early.

> choosing 2^13 as the threshold to decide which strategy to use

I would suggest to use the only slightly larger 10^4 as more 'round' number here, but this is no strong opinion.

> Additionally, I did not include 7Y4L in the pytest testing code because its CIF file is excessively large (>100MB), which would bloat the repository size.

This makes sense, still we probably want to include the new approach in the tests. Instead we could mock the threshold constant in the test, to force the usage for the dictionary approach even for smaller structures

cloverzizi:
@padix-key  Sure, you can set the threshold to 10^4. Feel free to update this PR to adjust the threshold and add the test code.

src/biotite/structure/io/pdbx/convert.py
```diff
@@ -677,6 +677,59 @@ def _find_matches(query_arrays, reference_arrays):
...
+def _find_matches_by_dict(query_arrays, reference_arrays):
+    """
+    For each index in the `query_arrays` find the indices in the
+    `reference_arrays` where all query values match the reference counterpart.
+    If no match is found for a query, the corresponding index is -1.
+    """
+    # Convert reference arrays to a dictionary for O(1) lookups
+    reference_dict = {}
+    unambiguously_keys = set()
+    for idx, col in enumerate(np.stack(reference_arrays, axis=-1)):
```
padix-key, line 689: Running `np.stack()` forces the reference_arrays to the same dtype, probably still requiring time consuming conversion to string and losing the type information. For the purpose required here, `zip()` should achieve the same, right?

Furthermore, you are iterating over the *rows* of the `atom_site` (reference) and `struct_conn` (query) categories instead of *columns*, if I am not mistaken.

```suggestion
    for idx, row in enumerate(zip(*reference_arrays)):
```

src/biotite/structure/io/pdbx/convert.py
```diff
@@ -677,6 +677,59 @@ def _find_matches(query_arrays, reference_arrays):
...
+            unambiguously_keys.add(ref_key)
+            continue
+        reference_dict[ref_key] = idx
+
+    match_indices = []
+    for query_idx, query_col in enumerate(np.stack(query_arrays, axis=-1)):
+        query_key = tuple(query_col)
+        occurrence = reference_dict.get(query_key, -1)
+
+        if occurrence == -1:
```
padix-key, line 701: As we do not use NumPy here, we are able to use the more idiomatic `None` here.
```suggestion
        occurrence = reference_dict.get(query_key)

        if occurrence is None:
```

src/biotite/structure/io/pdbx/convert.py
```diff
@@ -677,6 +677,59 @@ def _find_matches(query_arrays, reference_arrays):
...
 
+def _find_matches_by_dict(query_arrays, reference_arrays):
+    """
+    For each index in the `query_arrays` find the indices in the
+    `reference_arrays` where all query values match the reference counterpart.
+    If no match is found for a query, the corresponding index is -1.
+    """
+    # Convert reference arrays to a dictionary for O(1) lookups
+    reference_dict = {}
+    unambiguously_keys = set()
```
padix-key, line 688: The set actually collects the keys that *are* ambiguous.

```suggestion
    ambiguous_keys = set()
```

src/biotite/structure/io/pdbx/convert.py
```diff
@@ -677,6 +677,59 @@ def _find_matches(query_arrays, reference_arrays):
...
+
+def _find_matches(query_arrays, reference_arrays):
+    """
+    For each index in the `query_arrays` find the indices in the
+    `reference_arrays` where all query values match the reference counterpart.
+    If no match is found for a query, the corresponding index is -1.
+    """
+    #  it was observed that when the size exceeds 2**13 (8192)
+    #  the dict strategy becomes significantly faster than the dense array
+    #  and does not cause excessive memory usage.
```
padix-key, line 725: Could you link your comment in this PR with the nice benchmark?

padix-key (changes requested):
I also found the time to review your PR now, below are some minor suggestions.

cloverzizi:
> I also found the time to review your PR now, below are some minor suggestions.

Thanks, I've revised the code following your suggestions.

padix-key:
I added a test that monkeypatches the threshold, so the new addition is now tested.

I can also confirm the performance issues with `7Y4L`. In my tests the dict approach was about 100 times faster :+1:. However I found a different value for the intersection of the performances during my profiling: Even for files where 'the product' (i.e `N_query * N_ref`) was 100000, the dense approach was still 2x faster. I located the threshold between 1,000,000 (`4V6X`) and 30,000,000 (`5LQP`) on my machine. Here is the script

```python
from pyinstrument import Profiler
import time
import biotite.database.rcsb as rcsb
import numpy as np

PDB_ID = "5LQP"
N = 10
PROFILING_INTERVAL = 1e-3


file_path = rcsb.fetch(PDB_ID, "bcif", ".")

pdbx.convert.FIND_MATCHES_SWITCH_THRESHOLD = np.inf
profiler = Profiler(interval=PROFILING_INTERVAL)
profiler.start()
for _ in range(N):
    pdbx_file = pdbx.BinaryCIFFile.read(file_path)
    pdbx.get_structure(pdbx_file, model=1, include_bonds=True)
profiler.stop()
profiler.write_html("dense.html")

pdbx.convert.FIND_MATCHES_SWITCH_THRESHOLD = 0
profiler = Profiler(interval=PROFILING_INTERVAL)
profiler.start()
for _ in range(N):
    pdbx_file = pdbx.BinaryCIFFile.read(file_path)
    pdbx.get_structure(pdbx_file, model=1, include_bonds=True)
profiler.stop()
profiler.write_html("dict.html")
```

Could you share the script you performed your benchmark with?

cloverzizi:
@padix-key Apologies for the previous error in my test code where the actual input size for the query did not match the printed output! Below is my corrected test code along with the updated chart. I find a threshold of 3 million to be appropriate.

```
import random
import sys
import time

import numpy as np

import biotite.structure.io.pdbx.convert as cvt


def test(thres):
    cvt.FIND_MATCHES_SWITCH_THRES = thres
    sizes = []
    times = []
    mems = []
    n_col = 9
    for n_r_i in range(10, 20, 5):
        n_r = 2**n_r_i
        reference_arrays = [np.array([f"a{i}" for i in range(n_r)])] * n_col
        for n_q_i in range(5, 10):
            n_q = 2**n_q_i
            query_arrays = [
                np.array(random.choices(reference_arrays[0], k=n_q))
            ] * n_col

            mem_of_array = (
                sys.getsizeof(np.ones((n_r, n_q, n_col), dtype=bool)) / 1048576
            )  # in MB
            mems.append(mem_of_array)

            size = int(np.log2(n_r * n_q))
            sizes.append(size)

            s = time.time()
            cvt._find_matches(query_arrays, reference_arrays)
            times.append(time.time() - s)
    return times, sizes, mems


def draw_size_vs_memory(array_sizes, array_mems):
    plt.figure(figsize=(10, 6))
    plt.scatter(array_sizes, array_mems, color="blue", label="dense array")
    plt.xlabel("$N_{\\text{query}} * N_{\\text{ref}}$ ($2^x$)")
    plt.ylabel("Memory (KB)")
    plt.title("Size vs. Memory (dense array)")
    plt.savefig("size_vs_memory.png")


def draw_size_vs_time(array_sizes, array_times, dict_times):
    plt.figure(figsize=(10, 6))
    plt.scatter(array_sizes, array_times, color="blue", label="dense array")
    plt.scatter(array_sizes, dict_times, color="red", label="dict")
    plt.xlabel("$N_{\\text{query}} * N_{\\text{ref}}$ ($2^x$)")
    plt.ylabel("times (s)")
    plt.legend()
    plt.title("Size vs. Time")
    plt.savefig("size_vs_time.png")


array_times, array_sizes, array_mems = test(float("inf"))  # dense array
dict_times, dict_sizes, _array_mems = test(0)  # dict
draw_size_vs_memory(array_sizes, array_mems)
draw_size_vs_time(array_sizes, array_times, dict_times)
print("size (2**n)", array_sizes)
print("array times:", array_times)
print("dict times:", dict_times)

# 3,000,000
reference_arrays = [np.array([f"a{i}" for i in range(100000)])] * 9
query_arrays = [np.array(random.choices(reference_arrays[0], k=30))] * 9

cvt.FIND_MATCHES_SWITCH_THRES = float("inf")
s = time.time()
cvt._find_matches(query_arrays, reference_arrays)
print("3M dense array: ", time.time() - s)

cvt.FIND_MATCHES_SWITCH_THRES = 0
s = time.time()
cvt._find_matches(query_arrays, reference_arrays)
print("3M dict: ", time.time() - s)
```
size_vs_memory
size_vs_time

padix-key:
Thanks for the update. That matches my approximate findings :+1:.

The failing tests originates from changed doctest results due to a new RDKit version. I'll quickly commit the fixed doctest output

padix-key (approved):

padix-key:
@Croydon-Brixton Do you have any remaining concerns?

Croydon-Brixton:
All good from my side! Thank you for taking a detailed look @padix-key and thank you for the addition @cloverzizi !
As a side-note, I'll be somewhat slow to respond in the next 2 weeks as I'm travelling through India for holiday.

PR #770, opened 2025-03-07
speed up `spread_X_wise`

Croydon-Brixton:
This tiny PR speeds up all uses of `spread_X_wise` in the codebase by replacing the python for-loop through numpy's vectorized analog. For small structures the performance improvement is negligible, but for large structures or if this method is called many times (e.g. in ML pipelines) this gives a slight boost.

### DIFF
--- a/src/biotite/structure/segments.py
+++ b/src/biotite/structure/segments.py
@@ -92,12 +92,8 @@ def spread_segment_wise(starts, input_data):
         Segment-wise spread `input_data`.
         Length is the same as `array_length()` of `array`.
     """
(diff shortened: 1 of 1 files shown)


Croydon-Brixton:
Out of curiosity I also had a quick look in the speed diff. between slice-ing vs `np.diff` and it appears that unless the arrays are 100k+ elements long, slicing is actually faster (at least on my Mac).

Screenshot 2025-03-07 at 09 55 26

padix-key (approved):
Nice find!

PR #775, opened 2025-03-13
Bypass DSSP in tests and documentation

padix-key:
Due to the ongoing problems with Conda DSSP builds, this PR bypasses the usage of `DsspApp` for now:

- `dssp` is removed from Conda environment.
- DSSP tutorial is not run but displayed statically (like the *NCBI BLAST* tutorial does for the sake of run time).
- Affected examples use other means to get the secondary structure for now.


padix-key:
I accepted the performance regressions reported above, as they are caused be updating the Python version. This is also the reason why some benchmarks are significantly faster than before.

PR #780, opened 2025-03-31
Remove default values from docstrings

padix-key:
Currently some docstrings also repeat the default value for a parameter. This is not necessary, as the function signature containing this information is also shown in the API reference. However, this introduces the problem that the function signature and docstring may get out-of-sync, if default values change. Example: `point_number` parameter in `sasa()`.

Hence, this PR removes the repeated default values from docstrings.

### DIFF
--- a/src/biotite/application/viennarna/rnaplot.py
+++ b/src/biotite/application/viennarna/rnaplot.py
@@ -28,7 +28,7 @@ class RNAplotApp(LocalApp):
 
     Parameters
     ----------
-    dot_bracket : str, optiona
+    dot_bracket : str, optional
         The structure in dot bracket notation.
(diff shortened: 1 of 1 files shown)


src/biotite/application/viennarna/rnaplot.py
```diff
@@ -28,15 +28,15 @@ class RNAplotApp(LocalApp):
 
     Parameters
     ----------
-    dot_bracket : str, optional (default: None)
+    dot_bracket : str, optiona
```
cisert, line 31:
```suggestion
    dot_bracket : str, optional
```

padix-key: Thanks. Seems I was to overeager 😄

Issue #777, opened 2025-03-27
Occasional missing inter-residue bonds with `altloc="occupancy"`


RafiBrent:
When calling `biotite.structure.io.pdbx.get_structure` with `altloc='occupancy'`,  there is an occasional bug where standard inter-residue polymer bonds are missing.

This happens because bonds are computed before filtering by altloc, but `biotite.structure.bonds._connect_inter_residue` will only use the first occurrence of a given `atom_name` when creating polymer bonds. Because of this, if the first-occurring version of an atom in the structure is not the highest-occupancy version, then when the first occurrence is filtered out the polymer bond is lost with it.

For a reproducing example, see PDB ID 3kz8:

```python
import numpy as np
from biotite.structure.io import pdbx

cif_file = pdbx.CIFFile.read("/path/to/3kz8.cif")
atom_array = pdbx.get_structure(cif_file.block, altloc="occupancy", use_author_fields=False, include_bonds=True)[0]

print(atom_array.atom_name[477], atom_array.res_name[477]) # This is a backbone carbonyl carbon atom in an amino acid
print(atom_array.atom_name[480], atom_array.res_name[480]) # This is a backbone nitrogen atom in an amino acid
print(atom_array.chain_id[477], atom_array.chain_id[480]) # They are in the same chain
bonded_idces = np.array(atom_array.bonds.get_bonds(477)[0]) 
print(bonded_idces) # However, they are not bonded!
```

(In the above example, there is also a missing inter-residue bond between atoms 482 and 488.)

Happy to provide any other information that would be helpful, and thanks so much for the fantastic library!

padix-key:
Hi, thanks for the thorough analysis! Your reasoning makes absolutely sense. I suspect this is easily fixed by moving the line

https://github.com/biotite-dev/biotite/blob/5465b750d843c11231845350e54c3bdbeb99055d/src/biotite/structure/io/pdbx/convert.py#L413

prior to the bond computation.

Would you like to create the fix, or should I prepare a PR? In case you tackle this issue: You can use `3o5r.cif` for testing this: This structure is already in the repo and has also altlocs in the protein chain.

RafiBrent:
[...]

PR #781, opened 2025-04-01
Fix: altloc bonds interaction

RafiBrent:
Corrects the bug addressed in #777 . This implementation is currently incomplete (a few pre-existing tests are failing).

### DIFF
--- a/src/biotite/structure/io/pdbx/convert.py
+++ b/src/biotite/structure/io/pdbx/convert.py
@@ -133,8 +133,7 @@ _other_type_list = [
 
 def _filter(category, index):
     """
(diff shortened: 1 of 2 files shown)


padix-key:
Thanks, I'll add the rest in the next days.

padix-key:
If found the problem: The inter-residue bonds were parsed from the complete `atom_site` category, so the atom indices of the `BondList` may point to the wrong atoms after altloc filtering. This is fixed by applying the altloc filter to the `atom_site` category too, before the bonds are parsed

RafiBrent:
Ah got it, that makes sense. Thanks for catching this!

PR #786, opened 2025-04-10
Check `sym_id` for determining residue and chain starts

padix-key:
`AtomArray` object from biological assemblies and unit cells set the `sym_id` annotation. This PR takes `sym_id` also into consideration when determining chain and residue starts. This is especially useful in the edge case, where a unit cell/assembly chain contains only a single residue/chain, as this still allows for distinguishing different molecules.

### DIFF
--- a/src/biotite/structure/box.py
+++ b/src/biotite/structure/box.py
@@ -272,6 +272,8 @@ def repeat_box(atoms, amount=1):
         The repeated atoms.
         Includes the original atoms (central box) in the beginning of
         the atom array (stack).
+        If the input contains the ``sym_id`` annotation, the IDs are continued in the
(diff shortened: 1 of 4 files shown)


davegrays:
nice! I was just about to open a PR for this, as I've run into the same issue.

Only in our case, we reconstruct the assemblies from the asym unit and provide a custom name for the annotation describing the symmetry ops.

Would it be possible to generalize this PR to accept a custom annotation instead of hardcoding it as "asym_id"?

padix-key:
I think I would still hardcode `sym_id`, as it is a standard annotation. However, I think it would be beneficial for edge cases such as yours to optionally support custom annotations as well. I will add this to this PR.

PR #804, opened 2025-05-15
Always write intra-residue bonds if available

padix-key:
With the addition of [NextGen PDBx files](https://www.wwpdb.org/ftp/pdb-nextgen-archive-site) the `include_bonds` parameter was previously added to `pdbx.set_structure()`. This option had the advantage that both the 'classical' PDBx file (without `chem_comp_bond` category) and 'next-gen' PDBx file (with `chem_comp_bond` category) could be written.

However, in the meantime the 'next-gen' files have become the new standard files in the PDB, i.e. downloaded PDBx files always contain the `chem_comp_bond` category. Hence, this parameter does not make sense anymore and in my opinion rather leads to confusion: the `struct_conn` category (inter-residue bonds) is always written, if bonds are available, but the `chem_comp_bond` category (intra-residue bonds) is only written if `include_bonds=True`.

In consequence, this PR deprecates the `include_bonds` parameter in `pdbx.set_structure()`. It does not have an effect anymore, and `chem_comp_bond` is always written if `AtomArray.bonds` is available.


padix-key:
The performance decreased in some benchmarks as the newly downloaded PDBx files have more categories than before.

PR #806, opened 2025-05-28
fix altloc parsing

kleinhenz:
Check that `altloc_ids.mask` is not `None` before accessing it.


padix-key (approved):
Looks good to me! The failing tests in `database.rcsb` will be fixed by #809. You can rebase on main, as soon as it is merged

padix-key:
I also just added the 3Di sequence for the new test structure, which was the reason for the previous test failure.

padix-key:
Yes, it is just an external database hickup, I just restarted the job. Can I merge as soon as it finishes?

PR #810, opened 2025-06-05
feat: find pi cation interactions

Jaakik:
# Add π-cation interaction detection

Adds `find_pi_cation_interactions()` function to `biotite.structure.rings` module for detecting interactions between aromatic rings and positively charged atoms.

- Follows same optimization patterns as `find_stacking_interactions` (vectorized operations, squared distances)
- Requires `charge` annotation to identify cations
- test with (3wip.cif) structure that includes cation-π interactions between the positively charged quaternary ammonium group of acetylcholine and aromatic residues in the binding pocket from Tryptophan and Tyrosine.

### DIFF
--- a/src/biotite/structure/rings.py
+++ b/src/biotite/structure/rings.py
@@ -326,8 +326,8 @@ def find_pi_cation_interactions(
     ...         structure.res_name[ring_indices[0]],
     ...         structure.res_name[cation_index]
     ...     )
-    TRP ACH
     TYR ACH
+    TRP ACH
     """
     if atoms.bonds is None:
         raise BadStructureError("Structure must have an associated BondList")


src/biotite/structure/rings.py
```diff
@@ -268,6 +273,122 @@ def find_stacking_interactions(
...
+
+        - The distance between ring centroid and cation must be within
+          `distance_cutoff`. Open Drug Discovery Toolkit (ODDT) uses 5.0 Å,
+          whereas Protein-Ligand Interaction Fingerprints (ProLIF) uses 4.5 Å.
+        - The angle between the ring plane normal and the centroid-cation
+          vector must be within `angle_tol` of 0° (perpendicular to plane).
+
+    Examples
+    --------
+    >>> structure = load_structure("protein_ligand.cif", include_bonds=True, extra_fields=["charge"])
```
padix-key, line 326: In the Biotite test suite the doctests are actually executed, so this line will fail the test, as `protein_ligand.cif` does not exist. However you can use the predefined `path_to_structures` variable that actually points to `tests/structure/data` to select an existing structure file. You can take an inspiration from this docstring:

https://github.com/biotite-dev/biotite/blob/5ac8154335f5de8b7952a9c4b55345d2cc1fa416/src/biotite/structure/rings.py#L172-L187

src/biotite/structure/rings.py
```diff
@@ -268,6 +273,122 @@ def find_stacking_interactions(
...
+    TRP ACH
+    TYR ACH
+    ```
+    """
+    if atoms.bonds is None:
+        raise BadStructureError("Structure must have an associated BondList")
+
+    if atoms.charge is None:
+        raise BadStructureError(
+            "Structure must have a 'charge' annotation to identify cations. "
```
padix-key, line 342:
```suggestion
            "Structure must have a 'charge' annotation to identify cations"
```

tests/structure/test_rings.py
```diff
@@ -188,3 +188,43 @@ def test_no_adjacent_stacking_interactions():
...
     assert len(interactions) == 0
+
+
+def test_find_pi_cation_interactions():
+    """
+    Test π-cation interaction detection between aromatic residues and charged ligands.
+    Uses PDB 3wip known to have pi-cation interactions between tryptophan/tyrosine
+    residues and acetylcholine (ACH) ligand.
+    """
+    pdbx_file = pdbx.CIFFile.read(Path(data_dir("structure")) / "3wip.cif")
```
padix-key, line 199: Could you perform a quick check if any of the other test structures already fulfills this purpose, if it is not too time-consuming? I am a bit reluctant to add new test structures, if they are not necessary, as every one increases the size of the repo. Or is there another special reason to pick `3wip`?

padix-key (commented):
Hi, thanks for the feature! I left a few comments below. In addition, could you add this function to `doc/apidoc.json` so it is findable in the API reference?

tests/structure/test_rings.py
Jaakik: image

Jaakik: Usually 1 out of 4 TRP would  be involved in pi-cation interactions, I tried looking for backbone pi cation interactions between aromatic residues and cationic residues. To check this further I ran a plip batched script to analyse the protein-ligands interactions and none of them yielded any pi-cation interactions. 

The choice of `3wip` was just based on a suggestion from perplexity, there maybe smaller pdb files that contain this interaction I can check for those : ))

padix-key (approved):
Thanks for the changes! I just fixed a typo in the doctest and wait until the CI passes.

PR #815, opened 2025-07-20
Add pandas-style query functionality to `AtomArray` and `AtomArrayStack` objects

Croydon-Brixton:
This pull request introduces new query capabilities to the `AtomArray` class in the `biotite.structure` module, enabling users to filter, mask, and retrieve indices of atoms based on convenient pandas-like query expressions (https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.query.html).

### New query functionality in `AtomArray`:

* **`query` method**: Subsets the AtomArray(Stack) to atoms which satisfy the query expression  (e.g. `atom_array.query("(chain_id == 'A') & (atom_name == 'CA')`)
* **`mask` method**: Returns a True/False mask with True for atoms that match the query expression (e.g. `atom_array.mask("(chain_id == 'A') & (atom_name == 'CA')`)
* **`idxs` method**: Returns the indices of atoms that match a query expression (e.g. `atom_array.idxs("(chain_id == 'A') & (atom_name == 'CA')`)

**Pitch:**
This allows easy compound queries such as `atom_array.mask("(chain_id in ['A', 'Z']) & (res_name not in ['ALA', 'GLY']) | (b_factor > 0.3)")` which can be convenient for structural analysis in jupyter notebooks and for configuring generic, simple tasks via strings in configs (like the yaml configs in hydra).

Would love to hear your thoughts @padix-key  - do you think this would be useful enough to add as a feature in our main distribution?

### DIFF
--- a/tests/structure/test_query.py
+++ b/tests/structure/test_query.py
@@ -447,7 +447,9 @@ class TestCompatibilityWithExistingCode:
         assert idxs.tolist() == [1]
 
     def test_compound_queries2(self, atom_array):
-        atom_array.set_annotation("b_factor", np.array([10, 51, 93, 40, 50, 60, 70], dtype=np.float32))
+        atom_array.set_annotation(
+            "b_factor", np.array([10, 51, 93, 40, 50, 60, 70], dtype=np.float32)
+        )
         result = atom_array.query("(res_name in ['ALA', 'GLY']) & (b_factor > 50)")
         assert result.array_length() == 2
         assert result.res_name[0] == "ALA"


BradyAJohnston:
Thanks for the tag @padix-key you are certainly correct that I would love to see some kind of string selection implemented. I too like the current way of creating selections with `biotite` as string selections don't allow for code completion / hints and requires memorising a whole new syntax without any help from the code editor. 

On that note, the question around this kind of feature is always "what syntax to use?". It might make more sense to instead try and match the syntax of other programs (of most use to me personally would be MDAnalysis, but that's purely selfish). 

Always the option of going forward of forging your own path and creating your own syntax, but then there is _yet another_ selection syntax to try and remember when switching between software.

Excited with whatever comes though!

padix-key:
> Always the option of going forward of forging your own path and creating your own syntax, but then there is yet another selection syntax to try and remember when switching between software.

I would argue that the syntax use here is not a new one: It is actually the same as in `pandas` selection with annotation categories instead of columns. This makes it also quite close to the shape a classical fancy indexing would look like.

Example:

- Fancy indexing: `atoms[atoms.atom_name == 'CA']`
- Atom selection: `atoms.select("atom_name == 'CA'")`

t0mdavid-m:
Generally I agree with @padix-key in the sense that I am personally also not a big fan of these types of queries in general. However, if there is demand for queries I believe it makes sense to support them.

The added value could be greater if we used an established query language that is not captured by the existing fancy indexing. Thus, I would be more in favour of supporting a commonly used DSL like the ones in PyMol/mdanalysis.

The already existing integration with ammolite could be a plus for PyMols DSL.

PR #819, opened 2025-07-26
use isinstance() to support subclassing

youqingxiaozhua:
Hi biotite team,

Thanks for this great tool! I implemented a wrapper upon AtomArray to support some custom functions, but the current type() function doesn't support subclassing, which will cause an error. I changed it to isinstance() to handle this.

This is a minor change that improves the flexibility of the API. Let me know if you have any questions.

Best,

### DIFF
--- a/src/biotite/structure/atoms.py
+++ b/src/biotite/structure/atoms.py
@@ -1554,7 +1554,7 @@ def coord(item):
         Atom coordinates.
     """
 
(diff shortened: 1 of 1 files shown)


padix-key (approved):

padix-key:
Thanks for the fix! Quite some time ago I already replaced type equality checks with `isinstance()` checks, but I seemingly missed this one.

Issue #818, opened 2025-07-25
`connect_via_residue_names` adds unwanted bonds to ligand while reading pdb file


Emrys-Merlin:
I have a dataset of pdb files that each contain a protein binding a ligand. The ligands differ from file to file. In each file the ligand always has the residue name "LIG". I want to load these files with biotite including the bonds. Sometimes, the loaded ligand has more bonds than specified by the "CONECT" entries in the PDB file. I have a minimal example below (containing only a ligand).

I traced the problem to [this line](https://github.com/biotite-dev/biotite/blob/dc996a5239f7df35eeba6b90395b478c8891e900/src/biotite/structure/io/pdb/file.py#L547) in `biotite.structure.io.pdb.file.PDBFile.get_structure`. The line calls `connect_via_residue_names`, which then looks up residues by residue name. Unfortunately, "LIG" is a valid entry in the used CCD and the method adds bonds if the atom names (by chance, in my case) match pairs in the `bonds_in_residue("LIG")` dictionary. This behavior was unexpected to me and I was wondering if it is intentional. 

For my use case, it would be great if the bond lookup in the CCD were restricted to the amino acids of the peptide. However, I'm not sure if that is too restrictive. As a workaround, I checked which keys return an empty dictionary in `bonds_in_residue` (at the time of writing, e.g., "BLU") and I could refactor the pdbs to use these keys as residue names. At the same time, that does not feel very stable to me. I would be happy about any input how to tackle my use case. If I am using the library wrong, I would also be happy to hear about it (and how to do it right 🙈).

I could think of two of solutions, if the observed behavior is undesired:
1. A switch to disable the bond lookup for HETATM entries.
2. A guaranteed residue name (or maybe even a range) that will never be looked up and that can be found in the docs.

Thank you for the great work and library!

# Example
```python
from pathlib import Path
import biotite.structure.io.pdb as pdb


def main():
    path = Path("./minimal.pdb")
    pdb_file = pdb.PDBFile.read(path)
    structure = pdb.get_structure(pdb_file, model=1, include_bonds=True)
    bond_list = structure.bonds
    print(bond_list.as_array())

if __name__ == "__main__":
    main()
```

Example PDB file (stub) - please ignore the identical coordinates
```
HETATM  704  C7  LIG B 101     -17.432  24.497  -0.918  1.00 26.28           C  
HETATM  705  C8  LIG B 101     -17.432  24.497  -0.918  1.00 26.28           C  
[...]

PR #820, opened 2025-08-01
Do not lookup bonds for hetero residues

Emrys-Merlin:
The PR aims at resolving issue [818](https://github.com/biotite-dev/biotite/issues/818). The idea is to disable the bond lookup in the compound dictionary for hetero atoms. I tried to keep the changes minimal, but I'm not sure if that introduces too much special casing. I'm looking forward to your feedback.

My first, naive approach to globally disable the bond lookup did not work, because the bond parsing from .pdb files and .cif files works a bit differently, as far as I understand. While for pdbs `CONECT` entries are parsed as bonds directly and then extended by bonds found in the dictionary lookup (see [here](https://github.com/biotite-dev/biotite/blob/a541322e2a55f668e12e5a138d796cd3f096f286/src/biotite/structure/io/pdb/file.py#L545C1-L550C1)), for cifs, the whole `_chem_comp_bond` table is parsed as the compound dictionary and all the bonds (for both hetero and regular atoms) are looked up in it (see [here](https://github.com/biotite-dev/biotite/blob/a541322e2a55f668e12e5a138d796cd3f096f286/src/biotite/structure/io/pdbx/convert.py#L417C1-L427C88)). So, I introduced the `ignore_hetero` flag, which is `False` by default (preserving the current behavior) and I set to `True` for pdb parsing. Lastly, I introduced water special casing, because in the test file 1crr, the pdb does not contain `CONECT` entries for the water atoms, while the CIF `_chem_comp_bond` does. I am not sure how common that scenario really is or if it would be possible to adapt the test case.

### DIFF
--- a/src/biotite/structure/bonds.pyx
+++ b/src/biotite/structure/bonds.pyx
@@ -1648,8 +1648,7 @@ def connect_via_distances(atoms, dict distance_range=None, bint inter_residue=Tr
 
 
 def connect_via_residue_names(atoms, bint inter_residue=True,
-                              dict custom_bond_dict=None,
-                              bint ignore_hetero=False):
+                              dict custom_bond_dict=None):
     """
     connect_via_residue_names(atoms, inter_residue=True, custom_bond_dict=None)
 
@@ -1675,8 +1674,6 @@ def connect_via_residue_names(atoms, bint inter_residue=True,
         respective :class:`BondType` (represented as integer).
         If given, these bonds are used instead of the bonds read from
         ``components.cif``.
-    ignore_hetero : bool, optional
-        If true, no bonds are created for hetero atoms.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


padix-key (changes requested):
Thanks for tackling this :+1:. I think this can be solely fixed by changes in `biotite.structure.io.pdb` as `custom_bond_dict` already supports the required flexibility. This way we would automatically respect the differences in PDB and PDBx bond definitions.

We could use the approach from [here](https://www.biotite-python.org/latest/apidoc/biotite.structure.connect_via_residue_names.html) (see the Notes section). We could supply the `custom_bond_dict` only with `hetero=False` residues plus water. So in `PDBFile` instead of calling `connect_via_residue_names(array)` we could do something like this:

```python
custom_bond_dict = {
    custom_bond_dict[res_name]: bonds_in_residue(res_name)
    for res_name in itertools.chain(np.unique(array.res_name), ["HOH"])
}
connect_via_residue_names(array, custom_bond_dict=custom_bond_dict)
```

padix-key (approved):

PR #822, opened 2025-08-19
Fix typos in docstrings

Emrys-Merlin:


Emrys-Merlin:
Hi,
I was hoping to getting around to finding more typos, but, to be honest, I'm not sure when I'll find the time. I switched it to ready for review.

PR #829, opened 2025-09-12
Disallow redundant bonds in `BondList`

Linford24:


padix-key:
Sorry for the late response, I was sick for a few days. I think these two tutorial pages give a sufficient overview of Cython's memory views:

https://cython.readthedocs.io/en/latest/src/userguide/numpy_tutorial.html
https://cython.readthedocs.io/en/stable/src/userguide/memoryviews.html

If you look at `bonds.pyx` you will often encounter lines, like these:

```
cdef uint32[:,:] all_bonds_v = self._bonds
```

It tells Cython to convert `self._bonds` into a typed memoryview of `uint32` values, which can then be efficiently accessed in C-code, without expensive calls to the Python API. You can access the dimensions of the memoryview using `.shape` and use integer indices to access specific elements. Note that fancy indexing is not possible for these constructs

PR #847, opened 2025-11-25
Add `pocketeer` to list of extensions

cch1999:
Add `pocketeer` package to the list of extensions in the `biotite` docs:

https://www.biotite-python.org/latest/extensions.html


padix-key:
I just pushed a commit that makes the link target the documentation website instead of the GitHub repo for consistency with the other links on the page. If you are fine with this I'd merge the PR.

PR #850, opened 2025-12-30
Use positional argument in np.reshape to keep working with newer numpy

sdvillal:
[Since numpy version 2.4.0](https://github.com/numpy/numpy/releases/tag/v2.4.0), [```newshape``` is now called ```shape``` in ```np.reshape```](https://github.com/numpy/numpy/pull/29994).


sdvillal:
Thanks for the quick turnaround @padix-key, and happy new year!

PR #880, opened 2026-04-05
feat: add ClustalW .aln format support for sequence alignments

### LINKED ISSUE #774: how save protein block based multiple sequence alignment as .aln file (clustalw format)
Hi,
I wish to know how I can convert the following alignment file into .aln format and save it in the local directory?

# Perform a multiple sequence alignment of the PB sequences
matrix = align.SubstitutionMatrix.std_protein_blocks_matrix()
alignment, order, _, _ = align.align_multiple(
    structural_sequences, matrix, gap_penalty=(-500, -100), terminal_penalty=False
)
alignment = alignment[:, order.tolist()]

thanks and best,
Ansuman

haoyu-haoyu:
## Summary

Add support for reading and writing ClustalW `.aln` alignment files, following the existing FASTA alignment I/O pattern.

## New files

| File | Description |
|------|-------------|
| `src/biotite/sequence/io/clustal/__init__.py` | Package exports |
| `src/biotite/sequence/io/clustal/file.py` | `ClustalFile` class (TextFile + MutableMapping) |
| `src/biotite/sequence/io/clustal/convert.py` | `get_alignment()` / `set_alignment()` conversion functions |
| `tests/sequence/data/clustal.aln` | Single-block test data |
| `tests/sequence/data/clustal_multi.aln` | Multi-block test data |
| `tests/sequence/test_clustal.py` | 9 test cases |

### ClustalFile
- Extends `TextFile` and `MutableMapping` (same pattern as `FastaFile`)
- Parses CLUSTAL header, sequence blocks, and consensus lines
- Dict-like interface: `clustal_file["seq1"]` returns gapped sequence string
- Handles multi-block alignments by concatenating segments per sequence name

### Conversion functions
- `get_alignment(clustal_file, seq_type=None)` — auto-detects nucleotide/protein
- `set_alignment(clustal_file, alignment, seq_names, line_length=60)` — formats into blocks

### Edge cases handled
- Empty file → `InvalidFileError`
- Missing CLUSTAL header → `InvalidFileError`
- Single sequence → clear `ValueError` (alignments need ≥ 2 sequences)
- `line_length=0` → `ValueError`
- Reusing a `ClustalFile` object → entries cleared before writing
- Consensus lines (with `*`, `:`, `.`) correctly skipped during parsing

## Tests

9 tests covering: file reading, multi-block parsing, dict interface, alignment conversion, round-trip consistency (object-level and file I/O-level), name count mismatch error, and explicit sequence type parameter.

Closes #774

### DIFF
--- a/src/biotite/sequence/io/clustal/__init__.py
+++ b/src/biotite/sequence/io/clustal/__init__.py
@@ -15,7 +15,7 @@ getting/setting directly :class:`Alignment` objects, rather than strings.
 """
 
 __name__ = "biotite.sequence.io.clustal"
-__author__ = "Biotite contributors"
+__author__ = "haoyu"
 
 from .convert import *
 from .file import *
--- a/src/biotite/sequence/io/clustal/__init__.py
+++ b/src/biotite/sequence/io/clustal/__init__.py
@@ -1,0 +1,21 @@
+# This source code is part of the Biotite package and is distributed
+# under the 3-Clause BSD License. Please see 'LICENSE.rst' for further
+# information.
+
+"""
+This subpackage is used for reading and writing sequence alignment objects
+using the ClustalW format (``.aln``).
+
+This package contains the :class:`ClustalFile`, which provides a
+dictionary-like interface to ClustalW files, where sequence names are
+keys and gapped sequence strings are the corresponding values.
+
+Furthermore, the package contains convenience functions for
+getting/setting directly :class:`Alignment` objects, rather than strings.
+"""
+
+__name__ = "biotite.sequence.io.clustal"
+__author__ = "Biotite contributors"
+
+from .convert import *
+from .file import *
--- a/src/biotite/sequence/io/clustal/convert.py
+++ b/src/biotite/sequence/io/clustal/convert.py
@@ -35,6 +35,11 @@ def get_alignment(clustal_file, seq_type=None):
         The alignment from the :class:`ClustalFile`.
     """
     seq_strings = list(clustal_file.values())
+    if len(seq_strings) < 2:
+        raise ValueError(
+            f"ClustalW file contains {len(seq_strings)} sequence(s), "
+            f"but an alignment requires at least 2"
+        )
     return Alignment.from_strings(
         seq_strings,
         functools.partial(_convert_to_sequence, seq_type=seq_type),
@@ -59,13 +64,18 @@ def set_alignment(clustal_file, alignment, seq_names, line_length=60):
         The number of sequence characters per line in each block.
         Default is 60.
     """
+    if line_length < 1:
+        raise ValueError(
+            f"'line_length' must be at least 1, got {line_length}"
+        )
     gapped_seq_strings = alignment.get_gapped_sequences()
     if len(gapped_seq_strings) != len(seq_names):
         raise ValueError(
             f"Alignment has {len(gapped_seq_strings)} sequences, "
             f"but {len(seq_names)} names were given"
         )
-    # Store sequences in internal dict
+    # Clear any existing entries before writing new data
+    clustal_file._entries.clear()
     for name, seq_str in zip(seq_names, gapped_seq_strings):
         clustal_file._entries[name] = seq_str
     # Rebuild the text lines with the specified line length
--- a/src/biotite/sequence/io/clustal/convert.py
+++ b/src/biotite/sequence/io/clustal/convert.py
@@ -1,0 +1,113 @@
+# This source code is part of the Biotite package and is distributed
+# under the 3-Clause BSD License. Please see 'LICENSE.rst' for further
+# information.
+
+__name__ = "biotite.sequence.io.clustal"
+__author__ = "Biotite contributors"
+
+import functools
+import warnings
+from collections import OrderedDict
+from biotite.sequence.align.alignment import Alignment
+from biotite.sequence.alphabet import AlphabetError, LetterAlphabet
+from biotite.sequence.seqtypes import NucleotideSequence, ProteinSequence
+
+__all__ = ["get_alignment", "set_alignment"]
+
+
+def get_alignment(clustal_file, seq_type=None):
+    """
+    Get an alignment from a :class:`ClustalFile` instance.
+
+    Parameters
+    ----------
+    clustal_file : ClustalFile
+        The :class:`ClustalFile` to be accessed.
+    seq_type : type[Sequence], optional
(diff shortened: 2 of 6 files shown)


padix-key:
Thanks for adding the parser :+1:. I'll have a look 👀

src/biotite/sequence/io/clustal/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+This package contains the :class:`ClustalFile`, which provides a
+dictionary-like interface to ClustalW files, where sequence names are
+keys and gapped sequence strings are the corresponding values.
+
+Furthermore, the package contains convenience functions for
+getting/setting directly :class:`Alignment` objects, rather than strings.
+"""
+
+__name__ = "biotite.sequence.io.clustal"
+__author__ = "Biotite contributors"
```
padix-key: You can put your own name here 🙂

src/biotite/sequence/io/clustal/convert.py
```diff
@@ -0,0 +1,123 @@
+# This source code is part of the Biotite package and is distributed
+# under the 3-Clause BSD License. Please see 'LICENSE.rst' for further
+# information.
+
+__name__ = "biotite.sequence.io.clustal"
+__author__ = "Biotite contributors"
```
padix-key: See above

src/biotite/sequence/io/clustal/file.py
```diff
@@ -0,0 +1,162 @@
+# This source code is part of the Biotite package and is distributed
+# under the 3-Clause BSD License. Please see 'LICENSE.rst' for further
+# information.
+
+__name__ = "biotite.sequence.io.clustal"
+__author__ = "Biotite contributors"
```
padix-key: See above

src/biotite/sequence/io/clustal/file.py
```diff
@@ -0,0 +1,162 @@
...
+
+        self._entries = OrderedDict()
+        for line in self.lines[1:]:
+            stripped = line.strip()
+            # Skip blank lines
+            if len(stripped) == 0:
+                continue
+            # Skip consensus lines (lines that start with whitespace
+            # and contain only consensus characters)
+            if line[0] == " " and all(c in _CONSENSUS_CHARS for c in line):
```
padix-key: - Completely empty lines would also pass this condition, right?
- Do we need the extra `line[0] == " "` check?
- I assume a regex would be faster than iterating over each character

src/biotite/sequence/io/clustal/file.py
```diff
@@ -0,0 +1,162 @@
...
+                continue
+            # Sequence lines: name followed by whitespace and sequence
+            # The name must start at the beginning of the line
+            # (no leading whitespace)
+            if line[0] != " ":
+                parts = line.split()
+                if len(parts) >= 2:
```
padix-key: In which valid case would we have a non-empty/non-consensus line with less then two parts? Wouldn't the `else` case warrant an exception?

tests/sequence/test_clustal.py
```diff
@@ -0,0 +1,121 @@
...
+    clustal.set_alignment(
+        file2, alignment, seq_names=["seq1", "seq2", "seq3"]
+    )
+    file2.write(buffer)
+
+    # Read back from string buffer
+    buffer.seek(0)
+    file3 = clustal.ClustalFile.read(buffer)
+    alignment3 = clustal.get_alignment(file3)
+    assert str(alignment) == str(alignment3)
```
padix-key: You should be able to compare the `Alignment` objects directly with each other.

```suggestion
    assert alignment == alignment3
```

padix-key (commented):
I finally found the time for review, thanks for you patience. This looks really good :+1:, I just have a few minor comments.

I think you need to run `ruff` on your branch and rebase it on `main`, then the CI should pass as well.

haoyu-haoyu:
Thanks @padix-key! Plan for the follow-up:

1. **`__author__`** (`__init__.py`, `convert.py`, `file.py`) — changing all three to `"haoyu"`, matching how sibling packages like `biotite.sequence.io.fasta` sign the author line.

2. **`file.py:94` consensus detection** — right, `line[0] == " "` is redundant after the `all(c in _CONSENSUS_CHARS for c in line)` check, and a completely empty line would also pass the latter (it is only excluded by the preceding blank-line skip, not by this condition itself). Switching to a single compiled `re.fullmatch(r"[ *:.]+", line)` — faster, reads more directly, and doesn't hinge on the preceding skip.

3. **`file.py:101` `len(parts) >= 2`** — good catch. A non-empty, non-consensus line with fewer than two whitespace-separated parts is a malformed sequence line; replacing the silent drop with an up-front `raise InvalidFileError(...)` (same exception used by the header check).

4. **`test_clustal.py:98`** — applying your suggestion. Will also switch `test_round_trip` at line 76 to the same `alignment == alignment2` comparison for consistency.

5. **Rebase onto `main` + `ruff`** — will do.

Will push and ping when ready.

haoyu-haoyu:
Pushed `7e6fb6e` (force-push, branch was rebased onto current `main`):

1. **`__author__`** — set to `"haoyu"` in `__init__.py`, `convert.py`, `file.py`.
2. **Consensus detection** — replaced with a precompiled `re.compile(r"[ *:.]+")` and `fullmatch` call. Since `TextFile.read()` uses `splitlines()`, trailing `\n` is already stripped, so the regex matches the full stored line.
3. **Malformed sequence lines** — `ClustalFile._parse()` now raises `InvalidFileError("Malformed sequence line, expected a name and a sequence segment separated by whitespace, got '…'")` when a non-empty, non-consensus line has fewer than two whitespace-separated parts.
4. **Alignment equality** — both `test_round_trip` and `test_write_read_round_trip` now compare `Alignment` objects directly.

All 9 tests pass locally, `ruff format` and `ruff check` are clean.

Ready for another look @padix-key.

padix-key (approved):
Looks good to me :+1:. Regarding the `__author__`, we typically add the full name here, but if you prefer to add you first name/nickname this would be also OK. Let me know when this PR is ready to merge from your side.

haoyu-haoyu:
> Looks good to me 👍. Regarding the `__author__`, we typically add the full name here, but if you prefer to add you first name/nickname this would be also OK. Let me know when this PR is ready to merge from your side.

No problem! "haoyu" is my name just like my github name😂, I'm ready for merge this PR, If it looks all good on your side, Let's do it then

PR #884, opened 2026-04-14
fix(structure): reject atom-to-self bonds in BondList and chem_comp_bond writer

### LINKED ISSUE #812: Bond can be created of an atom to itself
Currently you can add a bond to `BondList`, that connects an atom with itself. This should never be possible.

Tangentially, `pdbx.set_structure()` does not check atom uniqueness when writing the `chem_comp_bond` category: If an `AtomArray` connects two atoms that are equal in all annotations relevant for `chem_comp_bond`, The resulting `chem_comp_bond` category will contain a self-bond for these two atoms.

haoyu-haoyu:
Fixes #812.

A \`BondList\` should encode chemistry, so a bond whose two atom indices refer to the same atom is meaningless. Today both the constructor and \`add_bond()\` accept self-bonds silently.

### Changes

- \`BondList.add_bond\` raises \`ValueError\` when the resolved positive indices are equal, including the case where one index is given positive and the other negative (e.g. \`4, -1\` with \`atom_count=5\` — both resolve to atom \`4\`).
- \`BondList.__init__\` performs the same check on the bulk-add path after the per-row sort, so building a \`BondList\` from a numpy array that contains a self-bond row raises instead of silently storing it.
- The companion concern from #812 — that \`pdbx.set_structure()\` could emit a self-bond into \`chem_comp_bond\` when two distinct atoms in an \`AtomArray\` share the same \`(res_name, atom_name)\` annotations — is also addressed: \`_set_intra_residue_bonds()\` now filters bond rows where the two \`atom_name\` annotations are equal before deduplication, and returns \`None\` if no rows remain.

### Tests

- \`test_invalid_creation\` extended with two self-bond cases (positive index pair, mixed positive/negative).
- New \`test_add_self_bond_rejected\` covers \`add_bond()\` for both index forms and verifies the list is unchanged after each rejected add.

### DIFF
--- a/src/biotite/structure/bonds.pyx
+++ b/src/biotite/structure/bonds.pyx
@@ -306,6 +306,13 @@ class BondList(Copyable):
                     "Input array containing bonds must be either of shape "
                     "(n,2) or (n,3)"
                 )
+            # After per-row sorting a self-bond appears as a row whose two
+            # atom indices are equal. These cannot represent valid chemistry,
+            # so reject them at the construction boundary.
+            if (self._bonds[:, 0] == self._bonds[:, 1]).any():
+                raise ValueError(
+                    "Input contains a bond from an atom to itself"
+                )
             self._remove_redundant_bonds()
             self._max_bonds_per_atom = self._get_max_bonds_per_atom()
 
@@ -939,6 +946,11 @@ class BondList(Copyable):
 
         cdef uint32 index1 = _to_positive_index(atom_index1, self._atom_count)
         cdef uint32 index2 = _to_positive_index(atom_index2, self._atom_count)
+        if index1 == index2:
+            raise ValueError(
+                f"Cannot create a bond from an atom to itself "
+                f"(atom index {index1})"
+            )
         _sort(&index1, &index2)
 
         cdef int i
--- a/src/biotite/structure/io/pdbx/convert.py
+++ b/src/biotite/structure/io/pdbx/convert.py
@@ -1179,6 +1179,19 @@ def _set_intra_residue_bonds(array, atom_site):
     # Take the residue name from the first atom index, as the residue
     # name is the same for both atoms, since we have only intra bonds
     comp_id = array.res_name[bond_array[:, 0]]
+    # Two distinct atom indices can share the same (res_name, atom_name)
+    # annotations — that bond would surface in chem_comp_bond as a
+    # self-bond on the chemical component, which is meaningless.
+    not_self_bond = atom_id_1 != atom_id_2
+    atom_id_1 = atom_id_1[not_self_bond]
+    atom_id_2 = atom_id_2[not_self_bond]
+    comp_id = comp_id[not_self_bond]
+    bond_array = bond_array[not_self_bond]
+    value_order = value_order[not_self_bond]
+    aromatic_flag = aromatic_flag[not_self_bond]
+    any_mask = any_mask[not_self_bond]
+    if len(bond_array) == 0:
+        return None
     _, unique_indices = np.unique(
         np.stack([comp_id, atom_id_1, atom_id_2], axis=-1), axis=0, return_index=True
     )
--- a/src/biotite/structure/io/pdbx/convert.py
+++ b/src/biotite/structure/io/pdbx/convert.py
@@ -1179,19 +1179,15 @@ def _set_intra_residue_bonds(array, atom_site):
     # Take the residue name from the first atom index, as the residue
     # name is the same for both atoms, since we have only intra bonds
     comp_id = array.res_name[bond_array[:, 0]]
-    # Two distinct atom indices can share the same (res_name, atom_name)
-    # annotations — that bond would surface in chem_comp_bond as a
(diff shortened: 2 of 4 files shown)


src/biotite/structure/io/pdbx/convert.py
```diff
@@ -1179,6 +1179,19 @@ def _set_intra_residue_bonds(array, atom_site):
...
+    not_self_bond = atom_id_1 != atom_id_2
+    atom_id_1 = atom_id_1[not_self_bond]
+    atom_id_2 = atom_id_2[not_self_bond]
+    comp_id = comp_id[not_self_bond]
+    bond_array = bond_array[not_self_bond]
+    value_order = value_order[not_self_bond]
+    aromatic_flag = aromatic_flag[not_self_bond]
+    any_mask = any_mask[not_self_bond]
+    if len(bond_array) == 0:
+        return None
```
padix-key: If we have a self-bond I think it would be better to raise a `BadStructureError`.

padix-key (commented):
Hi, thanks for fixing the bug! Regarding the fix for `bonds.pyx`, this is already fixed in the `BondList` rewrite in the `rust` branch (https://github.com/biotite-dev/biotite/pull/871), which I plan to merge for the next Biotite release. If the Rust port is not included in the next release I probably would give https://github.com/biotite-dev/biotite/pull/829 priority here, as it was created earlier.

haoyu-haoyu:
Thanks @padix-key — agreed, `BadStructureError` is the right signal here. The function already raises `BadStructureError` for the adjacent "empty res_name" / "empty atom_name" preconditions, so raising one more for ambiguous annotations keeps the behavior consistent.

Plan for the follow-up push:

- Replace the silent `not_self_bond` filter in `_set_intra_residue_bonds` with an up-front check that raises `BadStructureError("Structure contains bonded atoms sharing the same (res_name, atom_name) annotation, which cannot be written to chem_comp_bond without producing a self-bond")`.
- Drop the now-redundant `if len(bond_array) == 0: return None` guard that existed only to catch the post-filter empty case — the earlier `len(bond_array) == 0` check on the unfiltered `bond_array` still handles the genuine "no intra bonds" path.
- Add a focused test in `tests/structure/io/pdbx/` that builds an `AtomArray` with two distinct atoms sharing `(res_name, atom_name)` plus a bond between them, and asserts `set_structure()` raises `BadStructureError`.

Will push and ping when ready.

haoyu-haoyu:
Pushed `aeea708`:

- `_set_intra_residue_bonds()` now raises `BadStructureError` up-front when any bond row has `atom_id_1 == atom_id_2` after annotation lookup, instead of filtering the row out. Error message: *"Structure contains bonded atoms sharing the same `(res_name, atom_name)` annotations, which cannot be written to chem_comp_bond without producing a self-bond."* — mirrors the style of the existing empty-name preconditions.
- Dropped the post-filter `if len(bond_array) == 0: return None` guard (the pre-filter guard at the top of the function still covers the genuine "no intra bonds" path).
- Added `tests/structure/io/test_pdbx.py::test_set_structure_self_bond_raises` — 2-atom `AtomArray` with both atoms annotated `(ALA, CA)` plus a bond between them, asserting `pdbx.set_structure()` raises.

Ready for another look @padix-key.

src/biotite/structure/io/pdbx/convert.py
```diff
@@ -1179,6 +1179,15 @@ def _set_intra_residue_bonds(array, atom_site):
     # Take the residue name from the first atom index, as the residue
     # name is the same for both atoms, since we have only intra bonds
     comp_id = array.res_name[bond_array[:, 0]]
+    # Two distinct atom indices sharing the same (res_name, atom_name)
+    # annotations would surface in chem_comp_bond as a self-bond on the
+    # chemical component, which is structurally invalid.
+    if np.any(atom_id_1 == atom_id_2):
```
padix-key, line 1185: This condition as it is would also pass, if two different resides (different `comp_id`) have the same `res_name`, right?

src/biotite/structure/io/pdbx/convert.py
```diff
@@ -1179,6 +1179,15 @@ def _set_intra_residue_bonds(array, atom_site):
...
+        raise BadStructureError(
+            "Structure contains bonded atoms sharing the same "
+            "(res_name, atom_name) annotations, which cannot be "
+            "written to chem_comp_bond without producing a self-bond"
```
padix-key, line 1189: More concise message.

```suggestion
            f"Atom '{atom_id_1}' in residue '{res_name}' forms self-bond"
```

src/biotite/structure/bonds.pyx
```diff
@@ -306,6 +306,13 @@ class BondList(Copyable):
                     "Input array containing bonds must be either of shape "
                     "(n,2) or (n,3)"
                 )
+            # After per-row sorting a self-bond appears as a row whose two
```
padix-key, line 309: As explained above, I would move remove the `bond.pyx` part of this PR

PR #928, opened 2026-07-16
Add FastqFile.Offset enum

### LINKED ISSUE #923: Make `FastqFile` offset an `IntEnum`
Currently the `offset` parameter of `FastqFile` expects a string which internally is converted into an integer. Introducing a `FastqFile.Offset` as inner `IntEnum` as alternative would more clearly list the possible options and map directly to the corresponding integer offset value.

ehsanestaji:
## Summary

- Add a nested `FastqFile.Offset` `IntEnum` for the supported FASTQ quality-score encodings.
- Use the enum in the constructor, `read()`, `read_iter()`, and `write_iter()` type hints and documentation.
- Update `biotite.application.sra` to expose the same offset API.
- Continue accepting integer offsets while removing the legacy string aliases for Biotite v2.

## Motivation

The enum makes the supported FASTQ variants discoverable while mapping directly to their numeric ASCII offsets. Since this change targets Biotite v2, the legacy string aliases are removed rather than carried forward as compatibility behavior.

## Validation

- `pixi run -e test test tests/sequence/test_fastq.py -q` — 23 passed
- `pixi run lint`

Closes #923

### DIFF
--- a/tests/sequence/test_fastq.py
+++ b/tests/sequence/test_fastq.py
@@ -56,31 +56,36 @@ def test_conversion(chars_per_line):
 
 
 @pytest.mark.parametrize(
-    ("member_name", "legacy_name", "expected_value"),
+    ("member_name", "expected_value"),
     [
-        ("SANGER", "Sanger", 33),
-        ("SOLEXA", "Solexa", 64),
-        ("ILLUMINA_1_3", "Illumina-1.3", 64),
-        ("ILLUMINA_1_5", "Illumina-1.5", 64),
-        ("ILLUMINA_1_8", "Illumina-1.8", 33),
+        ("SANGER", 33),
+        ("SOLEXA", 64),
+        ("ILLUMINA_1_3", 64),
+        ("ILLUMINA_1_5", 64),
+        ("ILLUMINA_1_8", 33),
     ],
 )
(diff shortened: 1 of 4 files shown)


padix-key:
Thanks for tackling the issue! This change is planned for `v2` of Biotite (I also changed the target branch of this PR), so we do not need to keep the string values for compatibility reasons. Note that `biotite.application.sra` needs to be updated as well as it exposes the same `offset` values.

tests/sequence/test_fastq.py
```diff
@@ -55,6 +55,28 @@ def test_conversion(chars_per_line):
         assert np.array_equal(test_scores, ref_scores)
 
 
+@pytest.mark.parametrize(
```
padix-key, line 58: Without the need of the legacy names I think this test can be omitted.

padix-key (changes requested):

tests/sequence/test_fastq.py
ehsanestaji: Addressed in `eb4753b7` — I removed the obsolete legacy-string test. The broader update also applies `FastqFile.Offset` throughout `biotite.application.sra`. Validation: 23 FASTQ tests passed and `pixi run lint` completed successfully.

padix-key (approved):

PR #929, opened 2026-07-16
Set nonzero vina seed for determinism

jberg5:
Should fix this failing test:

```
=========================== short test summary info ============================
FAILED tests/application/test_autodock.py::test_docking[True] - assert np.float32(2.6669219) < 2.0
 +  where np.float32(2.6669219) = <function max at 0x7feb411b0f70>(array([0., 0., 0., ..., 0., 0., 0.], shape=(1776,), dtype=float32))
 +    where <function max at 0x7feb411b0f70> = np.max
 +    and   array([0., 0., 0., ..., 0., 0., 0.], shape=(1776,), dtype=float32) = <function distance at 0x7feb189d28e0>(array([[38.607, 14.202, 18.949],\n       [39.073, 14.187, 17.510],\n       [38.898, 15.623, 16.986],\n       ...,\n       ...49, 23.656,  8.504],\n       [49.002, 21.454,  6.825],\n       [50.497, 22.367,  7.110]], shape=(1776, 3), dtype=float32), array([\n	Atom(np.array([ 31.137,  21.711, -13.344], dtype=float32), chain_id="B", res_id=84, ins_code="", res_name="ARG", hetero=False, atom_name="CD", element="C", charge=0),\n	...,\n]))
 +      where <function distance at 0x7feb189d28e0> = struc.distance
= 1 failed, 943 passed, 7 skipped, 4 xpassed, 14 warnings in 157.23s (0:02:37) =
```

Reason for this is pretty wild, apparently Vina interprets `0` as "please give me a random seed"

https://github.com/ccsb-scripps/AutoDock-Vina/blob/3c65c0b3e6c2c1d183f6a175ecb65e3c5ba91645/src/lib/vina.cpp#L56

PR #933, opened 2026-07-30
fix dihedral_backbone() returning garbage angles across chain breaks

### LINKED ISSUE #744: Dihedral Backbone Does Not Account for Gaps, Returns Incorrect Values
I didn't see anything in the docs that says `biotite.structure.dihedral_backbone()` will return an incorrect value if there's a gap the protein backbone, i.e. it fails `biotite.structure.check_res_id_continuity()`.

For example in the (Ramachandran code sample)[https://www.biotite-python.org/latest/examples/gallery/structure/protein/ramachandran.html], with PDB: `3VKH` there are many gaps in that structure which would result in incorrect dihedrals at the boundaries when we pull atom coordinates between successive i and i+1 residues. In just chain A, there would be erroneous dihedrals produced at all the following positions:
```
Couldn't find next residue data for residue 1619
Couldn't find next residue data for residue 2060
Couldn't find next residue data for residue 2169
Couldn't find next residue data for residue 2452
Couldn't find next residue data for residue 2630
Couldn't find next residue data for residue 3211
Couldn't find next residue data for residue 3491
Couldn't find next residue data for residue 3736
Couldn't find next residue data for residue 4438
Couldn't find next residue data for residue 4465
Couldn't find next residue data for residue 4521
Couldn't find next residue data for residue 4581
Couldn't find next residue data for residue 4613
Couldn't find next residue data for residue 4624
Couldn't find next residue data for residue 4673
Couldn't find next residue data for residue 4729
```

Here's an example from the above PDB which isolates one of these gaps (with 2 residues on either side)
[3vkh_gap.pdb.txt](https://github.com/user-attachments/files/18582570/3vkh_gap.pdb.txt)
In this case, phi/psi/omega, in degrees, are:
```
(array([        nan,   42.100914,  -94.83789 , -118.06954 ], dtype=float32), 
array([156.29105 , -47.083096, 123.67293 ,        nan], dtype=float32), 
array([ 179.54808 ,   31.235714, -179.63321 ,         nan], dtype=float32))
```
The phi value `-94.837` and the psi/omega the values `-47.084` and `31.235` would be incorrect due to the unphysical gap, both would be better as `nan`.

----

For comparison, MDAnalysis returns `None` when you try to get dihedrals across a gap, for example:
```
import MDAnalysis as mda
from MDAnalysis.analysis import dihedrals
import numpy as np
u = mda.Universe("3vkh_gap.pdb")
protein = u.select_atoms('protein')
print('There are {} residues in the protein'.format(len(protein.residues)))
omegas = [res.omega_selection() for res in protein.residues]
for omega in omegas:
    if omega:
        print(omega, omega.dihedral.value())
```
returns the correct non-nan angles only:
```
There are 4 residues in the protein
<AtomGroup [<Atom 2: CA of type C of resname LEU, resid 2059 and segid A and altLoc >, <Atom 3: C of type C of resname LEU, resid 2059 and segid A and altLoc >, <Atom 9: N of type N of resname LEU, resid 2060 and segid A and altLoc >, <Atom 10: CA of type C of resname LEU, resid 2060 and segid A and altLoc >]> 179.54809477851654
<AtomGroup [<Atom 18: CA of type C of resname ASN, resid 2064 and segid A and altLoc >, <Atom 19: C of type C of resname ASN, resid 2064 and segid A and altLoc >, <Atom 22: N of type N of resname ILE, resid 2065 and segid A and altLoc >, <Atom 23: CA of type C of resname ILE, resid 2065 and segid A and altLoc >]> -179.63322828758015
```

steps-re:
fixes #744

`dihedral_backbone()` builds its phi/psi/omega coordinate arrays purely by positional adjacency between residues in the AtomArray. it never checks whether two consecutive residues are actually bonded. so if a structure has a missing loop, or you concatenate two unrelated fragments into one AtomArray, the dihedrals spanning that gap come out as finite garbage instead of NaN, exactly like padix-key called out in the issue thread.

went with the approach discussed in the thread: check the C-N distance between consecutive residues directly on the coordinates `dihedral_backbone()` already extracted, rather than calling `check_backbone_continuity()` separately (which would mean re-filtering the whole array again). used the same 1.2-1.8 A tolerance that `check_backbone_continuity()` / `filter_linear_bond_continuity()` already use elsewhere in the codebase, so the threshold is consistent across the library. where the C(i)-N(i+1) distance falls outside that range, `psi[i]`, `omega[i]`, and `phi[i+1]` get set to NaN since those are the three angles that straddle the junction between residue i and i+1.

while testing this i found `nucleotide_dihedral_backbone()` has the exact same bug for the O3'-P bond between residues, so I fixed that too with the same approach (`epsilon[i]`, `zeta[i]`, `alpha[i+1]` get NaN'd on a broken O3'(i)-P(i+1) junction). that one isn't filed anywhere, just the same root cause turning up twice.

verification:
- added `test_dihedral_backbone_chain_break` and `test_nucleotide_dihedral_backbone_chain_break` to `tests/structure/test_geometry.py`. both build a synthetic structure from two real fragments of 1l2y / 4p5j, renumber the second fragment's res_id to be perfectly continuous with the first (so the test can't accidentally pass just because of a res_id jump), then translate it 1000 A away so the two fragments are not bonded. confirmed both tests fail on current main (angles come back as finite numbers) and pass with this fix.
- ran the full `tests/structure/` suite (4176 passed, 13 skipped, no failures) to check for regressions, plus the existing multi-model, MDTraj-consistency, and barnaba-consistency dihedral tests specifically.
- ruff check + ruff format clean on the touched files.

disclosure: this fix (code + tests) was put together with AI assistance (Claude), then reviewed and verified by me before opening the PR.

### DIFF
--- a/tests/structure/test_geometry.py
+++ b/tests/structure/test_geometry.py
@@ -99,18 +99,39 @@ def test_dihedral_backbone_consistency(multi_model):
 
 
 @pytest.mark.parametrize("multi_model", [False, True])
-def test_dihedral_backbone_chain_break(multi_model):
+@pytest.mark.parametrize(
+    "function_name", ["dihedral_backbone", "nucleotide_dihedral_backbone"]
+)
+def test_dihedral_backbone_chain_break(function_name, multi_model):
     """
-    :func:`dihedral_backbone()` must not compute dihedral angles across a
-    chain break, i.e. two residues that are positionally adjacent in the
-    :class:`AtomArray` but not actually bonded.
+    :func:`dihedral_backbone()` and :func:`nucleotide_dihedral_backbone()`
+    must not compute dihedral angles across a chain break, i.e. two
+    residues that are positionally adjacent in the :class:`AtomArray` but
+    not actually bonded.
     This is simulated by taking two unrelated fragments of the same chain
     and translating one of them far away, while keeping the residue IDs
     perfectly continuous, to ensure the detection is based on the actual
     bond distance and not on residue numbering.
     """
-    pdbx_file = pdbx.BinaryCIFFile.read(data_dir("structure") / "pdb" / "1l2y.bcif")
+    if function_name == "dihedral_backbone":
+        pdb_id = "1l2y"
+        angle_names = ["phi", "psi", "omega"]
+        # The angles reaching into the following residue and the angle
+        # reaching back into the preceding one
+        trailing_names = ["psi", "omega"]
+        leading_name = "phi"
+    else:
+        pdb_id = "4p5j"
+        angle_names = ["alpha", "beta", "gamma", "delta", "epsilon", "zeta"]
+        trailing_names = ["epsilon", "zeta"]
+        leading_name = "alpha"
+
+    pdbx_file = pdbx.BinaryCIFFile.read(
+        data_dir("structure") / "pdb" / f"{pdb_id}.bcif"
+    )
     atoms = pdbx.get_structure(pdbx_file, model=1)
+    if function_name == "nucleotide_dihedral_backbone":
+        atoms = atoms[struc.filter_canonical_nucleotides(atoms)]
 
     fragment_1 = atoms[np.isin(atoms.res_id, [1, 2])].copy()
     fragment_2 = atoms[np.isin(atoms.res_id, [15, 16])].copy()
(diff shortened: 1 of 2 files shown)


tests/structure/test_geometry.py
```diff
@@ -98,6 +98,80 @@ def test_dihedral_backbone_consistency(multi_model):
     assert test_ome == pytest.approx(ref_omega, abs=1e-3, nan_ok=True)
 
 
+@pytest.mark.parametrize("multi_model", [False, True])
+def test_dihedral_backbone_chain_break(multi_model):
```
padix-key: The setup of both test function is basically the same, so could you maybe combine them into one function parametrized on `["dihedral_backbone", "nucleotide_dihedral_backbone"]`?

tests/structure/test_geometry.py
```diff
@@ -98,6 +98,80 @@ def test_dihedral_backbone_consistency(multi_model):
...
+    assert np.all(np.isnan(omega[..., 1]))
+    assert np.all(np.isnan(phi[..., 2]))
+    # All other angles within a fragment are unaffected and must remain
+    # finite
+    assert np.all(np.isfinite(phi[..., 1]))
+    assert np.all(np.isfinite(psi[..., 0]))
+    assert np.all(np.isfinite(omega[..., 0]))
+    assert np.all(np.isfinite(psi[..., 2]))
+    assert np.all(np.isfinite(omega[..., 2]))
+    assert np.all(np.isfinite(phi[..., 3]))
```
padix-key: The order is a bit odd here :wink:. I'd say we either sort by angle name or by index.

padix-key (changes requested):
Hi, sorry for the delayed response, I was busy the last few days.
This looks quite good, I would only clean up the tests a bit.

steps-re:
both done. it's one test now, parametrized on the two functions, with the pdb id and the angle names picked per case. the returned angles go into a dict keyed by name, so the asserts group per angle in ascending index order instead of hopping around.

full test_geometry.py green locally, 31 passed.

padix-key (approved):
Thanks a lot :+1:

steps-re:
The CodSpeed report here looks like runner noise rather than a real regression.

The three flagged benchmarks are all `benchmark_cell_list_compute_contacts`. This PR only touches `geometry.py` and `test_geometry.py`, and nothing in `benchmarks/` calls `dihedral_backbone` or `nucleotide_dihedral_backbone`, so no benchmark in the suite exercises the changed code.

The same pattern shows up elsewhere. #935 reports the same `cell_list_compute_contacts[MASK]` benchmark going 16.2 ms to 18.6 ms off the identical base, and #929 (setting a Vina seed) reports a 16% improvement in the k-mer benchmarks. Both carry the same "Different runtime environments detected" warning this one does.

For what it's worth, the added check is one `np.linalg.norm` over the residue count. Locally it costs about 11 us on a 1294-residue structure where `dihedral_backbone` itself takes 938 us, so roughly 1%.

Happy to acknowledge it on CodSpeed instead if you'd rather clear the check that way.

padix-key:
> The CodSpeed report here looks like runner noise rather than a real regression

Indeed, specifically these tests have a noisy run time, where I did not find the cause yet. I acknowledged this regression.

padix-key:
`pyright` still complains in the CI. Could you have another look?

steps-re:
Good catch. The remaining pyright failures were six `reportIndexIssue` errors (geometry.py lines 721-723 and 979-981), all `"__getitem__" method not defined on type "floating[Any]"`.

Root cause is the same phantom branch we already annotate on the return statements: `dihedral()` is typed `NDArray1 | NDArray2 | np.floating`, so `phi`/`psi`/`omg` (and `alpha`/`epsilon`/`zeta`) still carry the scalar `np.floating` member. The new in-place NaN masking indexes those results, and pyright rejects the subscript against the scalar branch, even though it is unreachable here (the inputs are always at least 2D).

I narrowed the indexed results with `np.asarray(...)`, which drops the scalar branch and leaves a plain `ndarray`. It is a runtime no-op on arrays (no copy), so the numbers are unchanged, and it keeps this as a real narrowing rather than another suppression. This mirrors how the identical `displacement()` union is handled before indexing in `rings.py`. The existing `# pyright: ignore[reportReturnType]` on the two return lines still applies, since the shaped `NDArray1`/`NDArray2` aliases don't match a bare `ndarray`.

Verified locally with the pinned `pyright == 1.1.409`: zero `reportIndexIssue` on `geometry.py`, and `pytest tests/structure/test_geometry.py -k dihedral` stays green (14 passed), including the chain-break cases.

PR #934, opened 2026-07-31
Added transformation return to orient_principal_components

nablacart:
Modified `transform.orient_principal_components` to return an `AffineTransformation` of centering translation and cumulative rotation to align with specified axes, in addition to transformed coordinates. Updated tests to reflect changes.

Returning the transformation allows for downstream alignment to a specific component of the input structure.

Validated with:
- `pixi run -e test test tests/structure/test_transform.py -q`: 1135 passed


src/biotite/structure/transform.py
```diff
@@ -630,9 +635,16 @@ def orient_principal_components(
         rotation = v @ wt
         if np.isclose(rotation, identity, atol=1e-5).all():
             break
+        cumulative_rotation = cumulative_rotation @ rotation
         # Apply rotation, keep molecule centered on the origin
         centered = centered @ rotation
```
padix-key, line 640: As we have a `AffineTransformation` at hand now, I see no reason to carry `centered` around. Instead we can simply

```
return transform.apply(atoms), transform
```

padix-key (changes requested):
I think this is a good idea :+1:. However it changes the function signature, so it would put this into the `biotite` `v2` release (which is already planned after the next release).

Otherwise, I left a minor review comment

padix-key:
I'll change the target branch to `v2`. Could you rebase your branch on it?

padix-key:
If I see correctly you merged previous PRs from `v2` into this one instead of rebasing, right? Could you instead run `git rebase -i v2` and only keep the commits relevant to this PR?

nablacart:
> If I see correctly you merged previous PRs from `v2` into this one instead of rebasing, right? Could you instead run `git rebase -i v2` and only keep the commits relevant to this PR?

My apologies, I am inexperienced with contributing from a fork instead of a branch. I did rebase on v2 but may have done so incorrectly, so I'll revert and try again.

padix-key:
No worries, if you prefer I can also rerun the rebase (although I am not sure if GitHub allows me force-pushing to your branch afterwards)

PR #936, opened 2026-08-07
Vectorize per-element Python loops in hbond(), sasa() and set_structure()

steps-re:
Four hot paths in `biotite.structure` loop in Python over every atom or bond, where the number of distinct values involved is only a handful. This replaces those loops with vectorized equivalents.

Measured on an antibody structure (1igy, 12750 atoms) unless noted, against `main`:

| | before | after | |
|---|---|---|---|
| `hbond()` (1gya, no `BondList`) | 2.47 ms | 0.44 ms | 5.7x |
| `hbond()` (no `BondList`) | 68.7 ms | 18.8 ms | 3.7x |
| `set_structure()` with bonds | 18.3 ms | 6.3 ms | 2.9x |
| `residue_iter()` | 16.1 ms | 12.2 ms | 1.32x |
| `base_pairs()` (4p5j) | 37.8 ms | 32.9 ms | 1.15x |
| `sasa()` (ProtOr) | 135 ms | 129 ms | 1.05x |

### What changed

- **`hbond()`** - `_get_bonded_h_via_distance()` iterated over every donor and built a mask over the whole atom array each time, so the cost scaled with donor count times atom count. The hydrogen atoms are now sorted by residue ID once, each donor's candidates are located by binary search, and all distances are computed in one call. Structures without hydrogen atoms previously paid the full cost for a guaranteed empty result.
- **`set_structure()`** - `_set_intra_residue_bonds()` translated the `BondType` of every bond individually. Writing 1igy meant 11710 calls for about five distinct types.
- **`sasa()`** - both radii sets were resolved with a per-atom lookup, although a structure has far fewer distinct residue/atom name combinations than atoms. This also removes the duplication between the `ProtOr` and `Single` branches.
- **`_subarray()`** - indexing an atom array constructed the new object through its constructor, which allocates the seven standard annotation arrays, then immediately replaced all of them. Bypassing the constructor helps code that slices repeatedly.
- A benchmark for `hbond()` is added, since neither code path was covered.

### One behaviour change

The `_subarray()` commit changes one thing beyond performance. If an annotation category had been removed via `del_annotation()`, indexing used to bring it back as a zero-length array while `array_length()` reported the indexed length:

```python
array.del_annotation("chain_id")
sub = array[1:4]
sub.array_length()   # 3
len(sub.chain_id)    # 0
repr(sub)            # IndexError
```

The category now simply stays absent. `test_indexing_after_annotation_deletion` covers this and fails without the fix.

### Verification

Full test suite passes. Output was also compared directly against `main` over all 23 structures in the test data: donor/hydrogen assignments, SASA for both radii sets, sliced arrays with their annotations and bonds, residue iteration, and the bytes of `set_structure()` round trips through both CIF and BinaryCIF. All 759 compared arrays are identical, including 16.2 MB of written file content, with and without a periodic box.

Happy to split this up if you would rather review the `atoms.py` change separately.

---

Disclosure: I use Claude Code to help with these contributions. Every change here was reproduced, benchmarked and tested by me before opening, and this description is my own.

### DIFF
--- a/tests/structure/test_atoms.py
+++ b/tests/structure/test_atoms.py
@@ -75,6 +75,20 @@ def test_access(array):
         array.set_annotation("test2", np.array([0, 1, 2, 3]))
 
 
+def test_indexing_after_annotation_deletion(array, stack):
+    """
+    Indexing must not resurrect a deleted annotation category, as the
+    resurrected array would not match the length of the indexed object.
+    """
+    array.del_annotation("chain_id")
+    stack.del_annotation("chain_id")
+
+    for indexed in (array[1:4], stack[:, 1:4]):
+        assert "chain_id" not in indexed.get_annotation_categories()
+        for category in indexed.get_annotation_categories():
+            assert len(indexed.get_annotation(category)) == indexed.array_length()
+
+
 def test_finding_compatible_dtype(array):
     """
     Check if a compatible dtype is selected, if the existing one is incompatible with
--- a/tests/structure/test_atoms.py
+++ b/tests/structure/test_atoms.py
@@ -79,6 +79,12 @@ def test_indexing_after_annotation_deletion(array, stack):
     """
     Indexing must not resurrect a deleted annotation category, as the
     resurrected array would not match the length of the indexed object.
+
+    Previously the indexed object was created via its constructor with length 0,
+    which already adds the default annotation categories, and only the categories
+    present in the original object were overwritten afterwards.
+    Hence a deleted default category reappeared in the indexed object as an
+    empty array.
     """
     array.del_annotation("chain_id")
     stack.del_annotation("chain_id")
--- a/benchmarks/structure/benchmark_hbond.py
+++ b/benchmarks/structure/benchmark_hbond.py
@@ -1,0 +1,41 @@
+import pytest
+import biotite.structure as struc
+import biotite.structure.io.pdbx as pdbx
+from tests.util import data_dir
+
+
+@pytest.fixture(scope="module")
+def atoms():
+    """
+    A structure that includes hydrogen atoms.
+    """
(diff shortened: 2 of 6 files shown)


tests/structure/test_atoms.py
```diff
@@ -75,6 +75,20 @@ def test_access(array):
         array.set_annotation("test2", np.array([0, 1, 2, 3]))
 
 
+def test_indexing_after_annotation_deletion(array, stack):
+    """
+    Indexing must not resurrect a deleted annotation category, as the
+    resurrected array would not match the length of the indexed object.
```
padix-key, line 81: Could you add the motivation here, i.e. why there indexing could resurrect a deleted annotation array in the first place?

padix-key:
Good idea to only run a Python loop over the unique elements of an array instead of every element :+1:.

I wonder if we can cast this general pattern into some function to avoid the addition of complexity in each of the improved functions. If you allow, I'd like to play around with this idea on top of your PR in the next few days 🙂 .

steps-re:
yeah go for it, i would be curious to see what the general version looks like. the per-function repetition bugged me too, i just did not want to scope-creep the PR into an api change.

steps-re:
ci died on `socket hang up` pulling pixi, twice, before lint or the ccd build ran. nothing to do with the diff. bouncing it to retrigger.

padix-key:
Thanks for your patience. I'll merge this PR as it is: `numpy >= 2.0` introduces `numpy.unique_inverse()` which I'd use for the generalized approach. Hence, I will build upon your work here in #942 which will be merged into the Biotite `v2` branch.

padix-key:
@steps-re FYI: This commit has my generalized take of your refactoring: https://github.com/biotite-dev/biotite/pull/942/changes/b0310f9ce7853c30baf7674d3d2ee9866458ea83

PR #938, opened 2026-09-03
Take the ProtOr element from the CCD, not from the atom name

dylanpulver:
`_calculate_protor_radii()` takes an atom's element from the first character of its atom name. Wrong when the element has two letters, or a hydrogen's name does not start with `H`.

On `main` @ `af1c4bb`:

```
vdw_radius_protor("MSE", "SE") -> 1.77  # sulfur's radius, for selenium (same for SEC)
vdw_radius_protor("PHG", "C4") -> 1.76  # mercury named 'HG' read as a bonded H
vdw_radius_protor("GLY", "CA") -> 1.88  # control, unaffected
```

Across the CCD, **262 atoms in 175 components** change: 187 had their own element misread (139 Se, 23 H, rest Cu/Ni/Ca/Cd/Co/Cr/Na/Cl read as C or N); 75 are heavy atoms with a wrong hydrogen count, because a neighbour named `HG`/`HF` counted as hydrogen or one named `1HZ`/`D1` did not. `sasa(vdw_radii="ProtOr")`, the default, calls this, using its 1.8 fallback where `None` is now returned.

The element now comes from `chem_comp_atom.type_symbol`, the same CCD entry the bonds come from. `vdw_radius_single()` already uses the annotated element; only this path re-derived it.

`test_protor_radii` cannot see it: it runs 1GYA filtered to standard amino acids — all one-letter elements — and only asserts the radius is not `None`.

`tests/structure` collects 4197: 4184 passed / 13 skipped; reverting only `radii.py` fails exactly the 4 new cases. A naive "two-letter prefix if it is an element" fix instead breaks the existing `test_protor_radii`, as `CA` reads as calcium.

Drafted with Claude Opus 5 (`claude-opus-5`); every measurement was run.


src/biotite/structure/info/radii.py
padix-key, line 198: I know it is not directly connected to this PR, but could you add a cache here? It would make consecutive lookups much faster
```suggestion
@functools.cache
def _calculate_protor_radii(res_name: str) -> dict[str, float | None]:
```

src/biotite/structure/info/radii.py
```diff
@@ -209,7 +215,7 @@ def _calculate_protor_radii(res_name: str) -> dict[str, float | None]:
         # One time the first atom is the one to get valency and H count
         # for and the other time vice versa
         for main_atom, bound_atom in ((atom1, atom2), (atom2, atom1)):
-            element = main_atom[0]
+            element = elements.get(main_atom, "")
```
padix-key, line 218: Is there a proper case where we would expect that the `elements` dictionary would miss an atom? If not, I would like to make sure a `KeyError` is raised if that happens.
```suggestion
            element = elements[main_atom]
```

src/biotite/structure/info/radii.py
```diff
@@ -225,14 +231,29 @@ def _calculate_protor_radii(res_name: str) -> dict[str, float | None]:
             # Increase valency by one, since the bond entry exists
             group[1] += 1
             # If the atom is bonded to hydrogen, increase H count
-            if bound_atom[0] == "H":
+            if elements.get(bound_atom, "") in ("H", "D"):
```
padix-key, line 234:
```suggestion
            if elements[bound_atom] in ("H", "D"):
```

src/biotite/structure/info/radii.py
```diff
@@ -225,14 +231,29 @@ def _calculate_protor_radii(res_name: str) -> dict[str, float | None]:
...
+    """
+    Map each atom name in the given residue to its element,
+    based on the *Chemical Component Dictionary*.
+    """
+    atom_names = get_from_ccd("chem_comp_atom", res_name, "atom_id")
+    elements = get_from_ccd("chem_comp_atom", res_name, "type_symbol")
+    if atom_names is None or elements is None:
+        return {}
+    return {
+        str(atom_name): str(element).upper()
```
padix-key, line 252: I think this is the more idiomatic (and likely faster) way to convert a numpy scalar into a native Python value.
```suggestion
        atom_name.item(): element.item().upper()
```

tests/structure/test_info.py
```diff
@@ -96,6 +96,36 @@ def test_protor_radii():
...
+        ("PHG", "C4", 1.61),
+        # Hydrogen atoms whose name starts with a digit (e.g. '1HZ')
+        # must still be counted as hydrogen
+        ("A1BZO", "CZ", 1.88),
+    ],
+)
+def test_protor_radii_element(res_name, atom_name, ref_radius):
+    """
+    Check that the element used for the ProtOr group is taken from the
+    *CCD*, not from the first character of the atom name.
```
padix-key, line 124: I would frame this test differently, as no code path remains that would use the first character.
```suggestion
    Check the correct ProtOr radii for manually checked examples.
```

padix-key (changes requested):
Thanks for fixing this oversight :+1:. I left a few comments below, otherwise this PR looks good to me!

padix-key (approved):

PR #939, opened 2026-09-03
Fix CTG assignment in genetic code tables 27-30

dylanpulver:
`codon_tables.txt` maps `CTG` to **Ala** in genetic codes 27-30 (Karyorelict, Condylostoma, Mesodinium, Peritrich Nuclear). NCBI maps it to **Leu**, so `translate()` silently returns a wrong amino acid:

```python
seq.NucleotideSequence("ATGCTGCTG").translate(
    complete=True, codon_table=seq.CodonTable.load(27))
# before: MAA          after: MLL
```

**This was never biotite's own choice.** The file is a faithful snapshot of NCBI `gc.prt` **v4.2**, added in Nov 2017 and untouched since. NCBI corrected it themselves in v4.3: *"Change to CTG -> Leu in genetic codes 27, 28, 29, 30."*

**Verification.** I diffed all 64 codons of all 25 tables against `gc.prt` (16,444 bytes, Version 4.6, sha256 `2aecbdd0…`): of 1,600 assignments, exactly these 4 differ. Confirmed independently against NCBI's Taxonomy page and Biopython. Structurally, codes 27-30 reassign only stop codons, yet their leucine box reads `LLLA`.

**Deliberately left out**, as a different class of change — happy to follow up if you want them: the `GTG` start missing from code 3 (gc.prt v4.4); code 24 named "Pterobranchia", which NCBI renamed "Rhabdopleuridae" (v4.6, and renaming would break `CodonTable.load("Pterobranchia Mitochondrial")`); codes 32/33 absent.

**Tests.** Table contents had no coverage: forcing `CTG -> Leu` in *every* table left the whole suite green. The new controls (3→Thr, 12→Ser, 26→Ala — correct, and unchanged here) catch that. 6053 passed, 132 skipped; ruff 0.9.7 clean.


tests/sequence/test_codon.py
```diff
@@ -39,6 +39,30 @@ def test_table_load(table_id):
...
+        (30, "L"),  # Peritrich Nuclear
+        # Controls: tables that genuinely reassign 'CTG'
+        (1, "L"),  # Standard
+        (3, "T"),  # Yeast Mitochondrial
+        (12, "S"),  # Alternative Yeast Nuclear
+        (26, "A"),  # Pachysolen tannophilus Nuclear
+        (31, "L"),  # Blastocrithidia Nuclear
+    ],
+)
+def test_ctg_assignment(table_id, expected_amino_acid):
```
padix-key, line 59: I think this test is not very meaningful: It just checks against manually looked up values from the codon table. This means it also would not have caught the wrong amino acid if we looked up the values from the existing wrong `gc.prt`. Hence, I suggest to remove this test.

padix-key (changes requested):
Thanks for the correction! I'd remove the test, otherwise it is good to merge.

padix-key (approved):

PR #946, opened 2026-09-17
Write `CONECT` records for bonds between residues differing in insertion code

### LINKED ISSUE #945: PDB writer drops bonds between residues that differ only in insertion code
When writing a PDB file, a bond between two residues that have the same `res_id` and `chain_id` but different `ins_code` gets no `CONECT` record, so it is lost on a PDB round trip. The typical case is a disulfide between e.g. `52` and `52A` in an antibody with Kabat/Chothia numbering.

`_remove_non_conect_bonds()` in `structure/io/pdb/file.py` decides whether a bond is inter-residue by comparing only `res_id` and `chain_id`:

```python
| (atoms.res_id[bond_array[:, 0]] != atoms.res_id[bond_array[:, 1]])
| (atoms.chain_id[bond_array[:, 0]] != atoms.chain_id[bond_array[:, 1]])
```

so `52`/`52A` counts as one residue and the bond is dropped as intra-residue. The PDBx writer uses `get_residue_starts_for()` in `_filter_bonds()` and keeps the bond.

```python
import io
import biotite
import biotite.structure as struc
import biotite.structure.io.pdb as pdb

# two cysteines that differ only in their insertion code, linked by a disulfide
atoms = struc.array([
    struc.Atom([0, 0, 0], chain_id="H", res_id=52, ins_code="", res_name="CYS", atom_name="CA", element="C"),
    struc.Atom([1, 0, 0], chain_id="H", res_id=52, ins_code="", res_name="CYS", atom_name="SG", element="S"),
    struc.Atom([5, 0, 0], chain_id="H", res_id=52, ins_code="A", res_name="CYS", atom_name="CA", element="C"),
    struc.Atom([3, 0, 0], chain_id="H", res_id=52, ins_code="A", res_name="CYS", atom_name="SG", element="S"),
])
atoms.bonds = struc.BondList(4, [[0, 1, 1], [2, 3, 1], [1, 3, 1]])

print("biotite", biotite.__version__)
f = pdb.PDBFile()
f.set_structure(atoms)
out = io.StringIO()
f.write(out)
print("CONECT lines:", [line for line in out.getvalue().splitlines() if line.startswith("CONECT")])
out.seek(0)
back = pdb.PDBFile.read(out).get_structure(model=1, include_bonds=True)
print("PDB round trip bonds: ", back.bonds.as_array().tolist())

c = pdbx.CIFFile()
pdbx.set_structure(c, atoms)
out = io.StringIO()
c.write(out)
out.seek(0)
back = pdbx.get_structure(pdbx.CIFFile.read(out), model=1, include_bonds=True)
print("CIF round trip bonds: ", sorted(back.bonds.as_array().tolist()))

# same disulfide, but second residue numbered 53 instead of 52A
atoms.res_id[2:] = 53
atoms.ins_code[2:] = ""
f = pdb.PDBFile()
f.set_structure(atoms)
out = io.StringIO()
f.write(out)
print("CONECT lines with res_id 52/53:", [line for line in out.getvalue().splitlines() if line.startswith("CONECT")])
```

Output on `main` (1aaf784a) and on 1.7.1:

```
CONECT lines: []
PDB round trip bonds:  []
CIF round trip bonds:  [[0, 1, 1], [1, 3, 1], [2, 3, 1]]
CONECT lines with res_id 52/53: ['CONECT    2    4 ...', 'CONECT    4    2 ...']
```

The two CA-SG bonds are intra-residue, so those not being written is expected. The SG-SG bond is the one that should be there.

I have a small fix that uses `get_residue_starts_for()` here as well, will open a PR.

haeganm:
Fixes #945

`_remove_non_conect_bonds()` found inter-residue bonds by comparing `res_id` and `chain_id` only. Two residues that differ just in `ins_code` (e.g. `52` and `52A`) looked like one residue, so a bond between them got no `CONECT` record and was lost when writing a PDB file.

This now compares residue starts via `get_residue_starts_for()`, the same way `_filter_bonds()` does for the PDBx writer.

Added one small test with a disulfide between `52` and `52A`. It fails on `main` and passes here. `tests/structure/io/test_pdb.py`, `test_pdbx.py` and `tests/structure/test_bonds.py` pass, `ruff` 0.9.7 and `pyright` are clean on the changed file.


tests/structure/io/test_pdb.py
```diff
@@ -599,6 +599,36 @@ def test_hetatm_intra_residue_bonds():
     np.testing.assert_array_equal(actual_bonds, expected_bonds)
 
 
+def test_inter_residue_bond_with_ins_code():
```
padix-key, line 602: Would it be possible to generalize this test? For example we could check that `PDBFile.set_structure()` adds the same `CONECT` records as there are in the PDB file downloaded from RCSB. The test assets do contain structures with insertion codes, so this should catch it. If for some reason complete `CONECT` equivalence is not possible, we could at least compare the number of CONECT records.

padix-key (changes requested):
Good catch! Thank you for the fix :+1:

haeganm:
Good idea. I replaced the synthetic test with one over every PDB test asset. It checks that the written `CONECT` records include all pairs from the original file's `CONECT` records (matched by chain, residue, insertion code and atom name, since serials get renumbered) and all inter-residue bonds of the parsed structure.

One thing I found while doing this: the RCSB comparison alone doesn't catch the bug. RCSB only writes `CONECT` for hetero groups and disulfides, and hetero bonds are kept by the separate hetero check. What the fix changes in `1igy` are 8 peptide bonds between Kabat-numbered residues like 82/82A, which RCSB doesn't record. So the inter-residue check is the half that fails on `main` (for `1igy` only) and passes with the fix. The RCSB check holds on both.

The failing "databases" job is the `biotite.database.rcsb` doctest, which is also failing on `main`.

tests/structure/io/test_pdb.py
```diff
@@ -599,6 +599,79 @@ def test_hetatm_intra_residue_bonds():
...
+)
+def test_conect_records(path):
+    """
+    Writing a structure should give ``CONECT`` records for all
+    ``CONECT`` records in the original PDB file and for all inter-residue
+    bonds, including bonds between residues that only differ in their
+    insertion code (e.g. in ``1igy``).
+    """
+    ref_file = pdb.PDBFile.read(path)
+    atoms = ref_file.get_structure(model=1, include_bonds=True)
```
padix-key: Are altlocs the reason the test cannot directly simply compare atom IDs from the `CONECT` records with each other? In this case you can use `altloc="all"`
```suggestion
    atoms = ref_file.get_structure(model=1, altloc="all", include_bonds=True)
```

padix-key (changes requested):

tests/structure/io/test_pdb.py
haeganm: Yes, altlocs and the renumbered serials. With `altloc="all"` and `extra_fields=["atom_id"]` the atom IDs can be compared directly, so I dropped the identity matching. Still fails on `main` for `1igy` only and passes with the fix.

tests/structure/io/test_pdb.py
```diff
@@ -599,6 +599,58 @@ def test_hetatm_intra_residue_bonds():
...
+    # Keep all atoms and their original IDs to be able to compare the
+    # atom IDs in the 'CONECT' records directly
+    atoms = ref_file.get_structure(
+        model=1, altloc="all", extra_fields=["atom_id"], include_bonds=True
+    )
+    test_file = pdb.PDBFile()
+    test_file.set_structure(atoms)
+    test_pairs = _conect_pairs(test_file)
+
+    assert _conect_pairs(ref_file) <= test_pairs
```
padix-key, line 639: Shouldn't these two sets be equal?

```suggestion
    assert _conect_pairs(ref_file) = test_pairs
```

